PR #26, opened 2020-08-07
[Eloquent] Reduce unnecessary dependencies

mjcarroll:
`ament_export_dependencies` was incorrectly exporting dependencies that are only used for test (and marked as such in `package.xml`).

This also removes some `set(CMAKE_BUILD_TYPE RELEASE)`, because this is actually handled by the build or packaging system.  For instance `colcon --cmake-args -DCMAKE_BUILD_TYPE=Release`

### DIFF
--- a/plansys2_bringup/CMakeLists.txt
+++ b/plansys2_bringup/CMakeLists.txt
@@ -2,8 +2,6 @@ project(plansys2_bringup)
 
 cmake_minimum_required(VERSION 3.5)
 
-set(CMAKE_BUILD_TYPE RELEASE)
-
 find_package(ament_cmake REQUIRED)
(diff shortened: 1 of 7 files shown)


mjcarroll:
@fmrico It looks like you have PR testing turned off, but this should clear up the errors you were seeing in the packaging builds.

fmrico:
Thanks, @mjcarroll !!

I will try to test what you said about PR Testing. I didn't notice this.

mjcarroll:
@fmrico If you could cut another release, that would be great.  I think this is the only regression holding up an eloquent sync.

fmrico:
Ok!! I will make the PR in a few minutes

PR #32, opened 2020-11-10
return a goal response even when rejected.

teyssieuman:
The return of a goal response was previously forgotten,
generating a compilation warning.

Signed-off-by: Fabrice Larribe <fabrice.larribe@safrangroup.com>


fmrico:
Thanks @teyssieuman !!

teyssieuman:
Thanks for your rapid answer!

PR #33, opened 2020-11-15
Merging the current master branch into foxy-devel

teyssieuman:


fmrico:
Hi @teyssieuman 

I am not sure if this is the best strategy, but I usually sync master and foxy-devel when creating the packages for each distro. This is a delicate step. I prefer to concentrate on the contributions to the master branch and merge from there.

Of course, I am open to other options, but I would need a justification for changing.

Thanks!!

PR #50, opened 2020-12-19
Improve test coverage

teyssieuman:


plansys2_domain_expert/include/plansys2_domain_expert/Types.hpp
```diff
@@ -104,7 +104,7 @@ class Assignment
    * @param[in] expression a bracketed e
    * @return a vector with the first level elements in the expression
    */
-  std::vector<std::string> splitExpr(const std::string & expression);
+  static std::vector<std::string> splitExpr(const std::string & expression);
```
fmrico, line 107: It is not necessary to convert it to `static` to check it. Create an `Assignment` object using the constructor with no parameters, and call the method (it's public).

plansys2_domain_expert/src/plansys2_domain_expert/Types.cpp
```diff
@@ -246,8 +246,10 @@ std::shared_ptr<TreeNode> get_tree_node(const std::string & expr)
         return pred;
       }
 
+    // LCOV_EXCL_START
```
fmrico, line 249: Why this?

plansys2_problem_expert/src/plansys2_problem_expert/ProblemExpert.cpp
```diff
@@ -434,8 +434,10 @@ ProblemExpert::checkPredicateTreeTypes(
       }
 
     default:
+      // LCOV_EXCL_START
```
fmrico, line 437: Why this?

plansys2_domain_expert/include/plansys2_domain_expert/Types.hpp
teyssieuman: Ok, but isn't it useful to have this function `static` since it do not manipulate any class member or method ?
It can be used as an utility function without creating an `assignment`. Is this another place to put it or another function that already have the same behavior ?

fmrico: Maybe we should extract it as an external function in the same file.

fmrico:
Well, I will merge as is. 

Thanks!!

PR #59, opened 2020-12-28
Fix the call of the get_problem_instance service

teyssieuman:
Attempt to fix the problem in the test case of the `get_problem_instance` problem.


plansys2_terminal/test/terminal_test.cpp
```diff
@@ -226,7 +235,12 @@ TEST(terminal_test, load_popf_plugin)
   }
 
   domain_node->trigger_transition(lifecycle_msgs::msg::Transition::TRANSITION_ACTIVATE);
-  problem_node->trigger_transition(lifecycle_msgs::msg::Transition::TRANSITION_ACTIVATE);
+
+  rclcpp_lifecycle::State state = problem_node->trigger_transition(
+    lifecycle_msgs::msg::Transition::TRANSITION_ACTIVATE);
+  std::cout << "problem_node->trigger_transition " << state.label() << std::endl;
```
fmrico, line 241: I prefer to minimize output. May we change this with an ASSERT_*?

teyssieuman: Ok this was debug trace that I just forgot to remove.
No problem to remove it...

teyssieuman, line 241:
```suggestion
```

fmrico:
Merging!!!

Thanks @teyssieuman

PR #64, opened 2021-01-06
Create tmp directories

teyssieuman:
This modification aims to create temp directories even of the `node_namespace` is compounded ( "/a/b/c") ...

### DIFF
--- a/plansys2_popf_plan_solver/src/plansys2_popf_plan_solver/popf_plan_solver.cpp
+++ b/plansys2_popf_plan_solver/src/plansys2_popf_plan_solver/popf_plan_solver.cpp
@@ -15,6 +15,7 @@
 #include <sys/stat.h>
 #include <sys/types.h>
 
+#include <filesystem>
 #include <string>
 #include <iostream>
 #include <cstdio>
@@ -36,10 +37,13 @@ POPFPlanSolver::getPlan(
   const std::string & node_namespace)
 {
   if (node_namespace != "") {
-    //  This doesn't work as cxx flags must apper at end of link options, and I didn't
-    //  find a way
-    // std::experimental::filesystem::create_directories("/tmp/" + node_namespace);
-    mkdir(("/tmp/" + node_namespace).c_str(), ACCESSPERMS);
+    std::filesystem::path tp = std::filesystem::temp_directory_path();
+    for (auto p : std::filesystem::path(node_namespace) ) {
+      if (p != std::filesystem::current_path().root_directory()) {
+        tp /= p;
+      }
+    }
+    std::filesystem::create_directories(tp);
   }
 
   Plan ret;


fmrico:
Hi @teyssieuman 

I don't fully agree with this change. I think it's positive for a user to trace where is the pddl files of each namespace. 

I would rather prefer something like: `mkdir(("/tmp/" + node_namespace+ " -p") ...`

fmrico:
I think `fs::create_directories` could do the job here

teyssieuman:
Hi @fmrico

This modification do not change the previous behavior that let the user able to trace where the pddl file is generated.
The `for` iterator iterates on the parts of the namespace.

I already use the `create_directories` function. This modifications aims to make possible the use of "compound" namespaces like (/a/b/c). I also took care of removing the first part of the pass if it correspond to the "root" because the concatenation of a path ("/tmp" for example) with another path containing a root, removes the first part of the path.

plansys2_popf_plan_solver/src/plansys2_popf_plan_solver/popf_plan_solver.cpp
```diff
@@ -36,10 +37,13 @@ POPFPlanSolver::getPlan(
   const std::string & node_namespace)
 {
   if (node_namespace != "") {
-    //  This doesn't work as cxx flags must apper at end of link options, and I didn't
-    //  find a way
-    // std::experimental::filesystem::create_directories("/tmp/" + node_namespace);
-    mkdir(("/tmp/" + node_namespace).c_str(), ACCESSPERMS);
+    std::filesystem::path tp = std::filesystem::temp_directory_path();
+    for (auto p : std::filesystem::path(node_namespace) ) {
```
teyssieuman, line 41: Iterate on the parts of the namespace :
"/a/b/c" --> "/", "a", "b", "c"

plansys2_popf_plan_solver/src/plansys2_popf_plan_solver/popf_plan_solver.cpp
```diff
@@ -36,10 +37,13 @@ POPFPlanSolver::getPlan(
...
+      if (p != std::filesystem::current_path().root_directory()) {
+        tp /= p;
+      }
```
teyssieuman, line 44: Only append the "namespace" part, if it do not corresponds to the root dirretory ("/").

teyssieuman:
Hi @fmrico 

Did you see this last comment?

Does it answer your expectations or is this anything else I can do?

fmrico:
Hi @teyssieuman 

I have just tested it, and it works like a charm. I had some doubts, but it comes from my doubts about how `std::filesystem::temp_directory_path` works. I confused it with `mktemp`.

Thanks for this valuable contribution :wink:

PR #72, opened 2021-01-15
Improved BTBuilder

fmrico:
Hi all,

This PR contains a new algorithm for building a BT from the plan. It is simpler than the previous one, and fix some issues like https://github.com/IntelligentRoboticsLabs/ros2_planning_system_examples/pull/12 https://github.com/IntelligentRoboticsLabs/ros2_planning_system/issues/67 and others...

Best


fmrico:
Now, the Behavior tree for the last plan is also  stored in `/tmp/{namespace}/bt.xml`

PR #79, opened 2021-01-23
Improve the information at runtime

fmrico:
Dear all,

This PR improves the information produced to monitor the execution:
1. While executing a plan, the Executor publish at 1Hz in `/action_execution_info` the feedback about every action while executing a plan. This message is a `plansys2_msgs::msg::ActionExecutionInfo`. 

```
int8 NOT_EXECUTED=1
int8 EXECUTING=2
int8 FAILED=3
int8 SUCCEEDED=4

int8 status

builtin_interfaces/Time start_stamp
builtin_interfaces/Time status_stamp

string action
string[] arguments

float32 completion
string message_status
```

2. The action performers publish at 1Hz a "heartbeat" at '/performers_status' (`plansys2_msgs::msg::ActionPerformerStatus`). You can subscribe to this topic, or check it with the PlanSys2 terminal command `check actors`:

```
int8 NOT_READY=0
int8 READY=1
int8 RUNNING=2
int8 FAILURE=4

int8 state
string action
string[] arguments
string[] specialized_arguments
string node_name
```

This info can be used to build tools to monitor the system at runtime, like the one proposed in https://github.com/IntelligentRoboticsLabs/ros2_planning_system/issues/74

I hope this helps!!
Signed-off-by: Francisco Martin Rico <fmrico@gmail.com>

### DIFF
--- a/plansys2_msgs/msg/Knownledge.msg
+++ b/plansys2_msgs/msg/Knownledge.msg
@@ -1,0 +1,3 @@
+string[] instances
+string[] predicates
+string goal
--- a/plansys2_bt_actions/test/unit/bt_action_test.cpp
+++ b/plansys2_bt_actions/test/unit/bt_action_test.cpp
@@ -29,7 +29,6 @@
 #include "plansys2_executor/ActionExecutor.hpp"
 
 #include "test_msgs/action/fibonacci.hpp"
-#include "plansys2_msgs/action/execute_action.hpp"
 #include "ament_index_cpp/get_package_share_directory.hpp"
 
 #include "rclcpp/rclcpp.hpp"
--- a/plansys2_executor/include/plansys2_executor/ActionExecutorClient.hpp
+++ b/plansys2_executor/include/plansys2_executor/ActionExecutorClient.hpp
@@ -20,6 +20,7 @@
 #include <vector>
 
 #include "plansys2_msgs/msg/action_execution.hpp"
+#include "plansys2_msgs/msg/action_performer_status.hpp"
 
 #include "plansys2_domain_expert/DomainExpertClient.hpp"
 #include "plansys2_problem_expert/ProblemExpertClient.hpp"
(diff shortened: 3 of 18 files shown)


fmrico:
Also, Plansys2 is publishing (with transient local QoS) in `/problem_expert/knownledge` (` plansys2_msgs/msg/Knownledge`) the information contained in the problem expert any time it is updated:
```
string[] instances
string[] predicates
string goal
```

plansys2_msgs/msg/Knownledge.msg
```diff
@@ -0,0 +1,3 @@
+string[] instances
+string[] predicates
+string goal
```
teyssieuman, line 3: @fmrico 
Isn't the `functions` valuations part of the knowledge ?

teyssieuman (commented):
I thing that the functions a missing from the "knowledwe". Is this on purpose ?

plansys2_msgs/msg/Knownledge.msg
fmrico: Yes, absolutely. Please, add it too.

Maybe we should add an example in the examples repo. I didn't add functions because the others elements were not in those examples.
Anyway, it was available, but not in plansys2 namespace.

Thanks to be aware of this.

teyssieuman: Ok, I just update to this the current version of the master, and the functions do not seems to be functional anymore. I try to figure out what append and I will create an issue.

fmrico: :thinking: I merged #82 , which included a lot of improvements. I ran the tests and tested https://github.com/IntelligentRoboticsLabs/ros2_planning_system_examples, and everything seemed to be working.  I think that this PR could affect this. Did you add tests for `functions`? I remember that there were some of them...

teyssieuman: In my application, I had issues that I finally manage to solve:
- The name of the service to manage function changed, but I found it quite easily
- A service "Set function" and a service "update function" have been created, and it took me some time to understand that the "update function" fails if the function have not been "set" before. Is this useful maintain two different service ?
- A last point is described in #87, and took me some time to understand what append.

I still have some performance issue in launching the execution of the plan. The start of plan execution take much longer than before. I'm currently trying to identify where those slowness came from.

fmrico: Did you find the performance issue?

PR #82, opened 2021-01-29
Add support for numeric conditions and effects.

jjzapf:


fmrico:
Hi @jjzapf 

Awesome work!! Thank you very much for this contribution.

I have tested it with t[he examples](https://github.com/IntelligentRoboticsLabs/ros2_planning_system_examples) and everything works well, with a little detail. The `Predicate` class is exposed to the user (for example, [here](https://github.com/IntelligentRoboticsLabs/ros2_planning_system_examples/blob/2fe89f82963cc860cd2d77752cec4a0bffc56468/plansys2_patrol_navigation_example/src/patrolling_controller_node.cpp#L43)) , and write `parser::pddl::tree::Predicate` is not very convenient. I prefer to maintain `plansys2::Predicate`. I propose you to make an alias by adding these lines at the final of Tree.h:

```
 namespace plansys2 {
  typedef parser::pddl::tree::Predicate Predicate;
  typedef parser::pddl::tree::Instance Instance;
  typedef parser::pddl::tree::Goal Goal;
}

#endif  // PLANSYS2_PDDL_PARSER__TREE_H_
```

What do you think?

jjzapf:
> I propose you to make an alias by adding these lines at the final of Tree.h:
> 
> ```
>  namespace plansys2 {
>   typedef parser::pddl::tree::Predicate Predicate;
>   typedef parser::pddl::tree::Instance Instance;
>   typedef parser::pddl::tree::Goal Goal;
> }
> 
> #endif  // PLANSYS2_PDDL_PARSER__TREE_H_
> ```
> 
> What do you think?

Sounds good to me. I will make this change and push this afternoon. Thanks for your quick review of the proposed changes!

jjzapf:
I pushed the requested change to Tree.h and tested with one of our controller nodes. I think your intent was to only use the aliases for user accessible PDDL parser constructs, e.g., from a user defined controller node. I therefore did not replace any of the `parser::pddl::tree` namespaces within the internal plansys2 packages.

PR #84, opened 2021-02-02
Moving publisher on_activate call to the ExecutorNode::on_activate ca…

xydesa:
…llback.

Signed-off-by: Alexander Xydes <alexander.xydes@navy.mil>

Based on the below ROS 2 demo, I think activating the publisher should be done in the on_activate callback inside ExecuteNode.

https://github.com/ros2/demos/blob/master/lifecycle/src/lifecycle_talker.cpp#L152


fmrico:
Thanks!! @xydesa

Issue #87, opened 2021-02-08
`and` not supported in `durative-action` conditions or effects.


teyssieuman:
Considering the two following durative actions:
```pddl
(:durative-action A
  :parameters ( )
  :duration (= ?duration 1)
  :conditions ( and
    (at start <predicate_1>)
    (at start <predicate_2>)
  )
  :effect (and
    (at end <predicate_3>)
    (at end <predicate_4>)
) 


(:durative-action B
  :parameters ( )
  :duration (= ?duration 1)
  :conditions ( and
    (at start (and ( <predicate_1> <predicate_2> ) ) )
  )
  :effect (and
    (at end (and ( <predicate_3> <predicate_4> ) ) )
) 
```

The durative actions `A` and `B` are accepted by POPF witch find a plan.

But the `B` action is not working correctly withe `plansys2`  :
- In `BTBuilder::get_base` the `parser::pddl::tree::AND` node is not supported in the `FUNCTION` case.

fmrico:
Hi @teyssieuman 

Does this bug still exist after the last refactors?

fmrico:
Hi @teyssieuman 

I am closing this issue. Pease, reopen it if this effect appears again in the last version.

Thanks!!

PR #88, opened 2021-02-11
utils-bug: Fixing bug in OR case of evaluate function in plansys2_exe…

jjzapf:
We discovered a bug in the OR case of the evaluate function in Utils.cpp. When aggregating truth values using an OR operator, the initial value should be false. As a side note, we are working on a unit test for Utils.cpp and hope to create another PR soon.


fmrico:
Hi @jjzapf 

Do you think that this could have a relation with #87 ?

teyssieuman:
@fmrico I don't thing so ... The #87 is more about a non-supported syntax.

jjzapf:
I believe @teyssieuman is correct that this is unrelated. I have not been able to look into that issue yet. I would venture to guess that some changes may be needed in the domain expert or possibly even the pddl parser to support the alternative syntax.

fmrico:
Thanks!! @jjzapf 

Yes, unit tests for Utils.cpp are very valuables :)

Issue #89, opened 2021-02-16
Cancelling a plan


rossj13:
I can't seem to see how I would cancel/ clear the problem client and executor client part way through executing a plan so I could make a new plan to execute. Am I missing something or is that something that can be added in future instances.

fmrico:
Hi @rossj13 

Just this week I am working on this. Stay tuned.

tdenewiler:
I'm having similar issues at the moment. I'd like to track any changes if it would help. I can build your development branch and test with my own code to provide feedback.

fmrico:
Hi @tdenewiler 

I have been working on this all week. I have fixed it, but I am adding a lot of tests to be sure that everything will be working. A think that this weekend I will do a PR. Maybe you could test/review it before merging when I create it.

tdenewiler:
I'll do my best to monitor your PRs and will let you know if I am able to test it this weekend. Thanks!

fmrico:
Hi @rossj13 @tdenewiler 

I am closing this, as canceling plans are available from #92 

Please, open it again if you find any issue

PR #92, opened 2021-02-21
Action execution refactoring

fmrico:
Hi all,

This week I have been working in a refactor of the executor. This solves #89 but also fixes other implementation problems.

The most important change in the API of the `ExecutoClient`. Now we have three new methods:

* `bool start_plan_execution()`: Start a execution of a new plan. It returns false if there is any problem with starting this new execution.
* `bool execute_and_check_plan()`: This method _must_ be called iteratively to refresh the result and feedback. It returns true if the plan is still executing.
* `void cancel_plan_execution()`: cancel the current plan execution.

It is a big change. I noticed that the executor was not ready to cancel the plan, and I had to include this option in messages, returned values...

Please, check it and discuss any concerns about it.

Best

Signed-off-by: Francisco Martin Rico <fmrico@gmail.com>

### DIFF
--- a/plansys2_bt_actions/src/plansys2_bt_actions/BTAction.cpp
+++ b/plansys2_bt_actions/src/plansys2_bt_actions/BTAction.cpp
@@ -47,18 +47,15 @@ BTAction::on_configure(const rclcpp_lifecycle::State & previous_state)
     RCLCPP_INFO_STREAM(get_logger(), "plugin: [" << plugin << "]");
   }
 
-  BT::BehaviorTreeFactory factory;
   BT::SharedLibrary loader;
 
   for (auto plugin : plugin_lib_names) {
-    factory.registerFromPlugin(loader.getOSName(plugin));
+    factory_.registerFromPlugin(loader.getOSName(plugin));
   }
 
   auto node = rclcpp::Node::make_shared(std::string(get_name()) + "_aux");
   blackboard_ = BT::Blackboard::create();
   blackboard_->set("node", node);
-
-  tree_ = factory.createTreeFromFile(bt_xml_file_, blackboard_);
 
   return ActionExecutorClient::on_configure(previous_state);
 }
@@ -66,29 +63,44 @@ BTAction::on_configure(const rclcpp_lifecycle::State & previous_state)
 rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn
 BTAction::on_activate(const rclcpp_lifecycle::State & previous_state)
 {
+  tree_ = factory_.createTreeFromFile(bt_xml_file_, blackboard_);
+
   for (int i = 0; i < get_arguments().size(); i++) {
     std::string argname = "arg" + std::to_string(i);
     blackboard_->set(argname, get_arguments()[i]);
   }
 
+  finished_ = false;
   return ActionExecutorClient::on_activate(previous_state);
+}
+
+rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn
+BTAction::on_deactivate(const rclcpp_lifecycle::State & previous_state)
+{
+  tree_.haltTree();
+
+  return ActionExecutorClient::on_deactivate(previous_state);
 }
 
 void
 BTAction::do_work()
 {
-  auto result = tree_.rootNode()->executeTick();
+  if (!finished_) {
+    auto result = tree_.rootNode()->executeTick();
 
-  switch (result) {
-    case BT::NodeStatus::SUCCESS:
-      finish(true, 1.0, "Action completed");
-      break;
-    case BT::NodeStatus::RUNNING:
-      send_feedback(0.0, "Action running");
-      break;
-    case BT::NodeStatus::FAILURE:
-      finish(false, 1.0, "Action failed");
-      break;
+    switch (result) {
+      case BT::NodeStatus::SUCCESS:
+        finish(true, 1.0, "Action completed");
+        finished_ = true;
+        break;
+      case BT::NodeStatus::RUNNING:
+        send_feedback(0.0, "Action running");
+        break;
+      case BT::NodeStatus::FAILURE:
+        finish(false, 1.0, "Action failed");
+        finished_ = true;
+        break;
+    }
   }
 }
 
--- a/plansys2_executor/src/plansys2_executor/ActionExecutor.cpp
+++ b/plansys2_executor/src/plansys2_executor/ActionExecutor.cpp
@@ -52,12 +52,14 @@ ActionExecutor::action_hub_callback(const plansys2_msgs::msg::ActionExecution::S
     case plansys2_msgs::msg::ActionExecution::REQUEST:
     case plansys2_msgs::msg::ActionExecution::CONFIRM:
     case plansys2_msgs::msg::ActionExecution::REJECT:
+    case plansys2_msgs::msg::ActionExecution::CANCEL:
       // These cases have no meaning requester
       break;
     case plansys2_msgs::msg::ActionExecution::RESPONSE:
       if (msg->arguments == action_params_ && msg->action == action_name_) {
         if (state_ == DEALING) {
           confirm_performer(msg->node_id);
+          current_performer_id_ = msg->node_id;
           state_ = RUNNING;
           waiting_timer_ = nullptr;
           start_execution_ = node_->now();
@@ -199,12 +201,26 @@ ActionExecutor::tick(const rclcpp::Time & now)
       break;
     case SUCCESS:
     case FAILURE:
+    case CANCELLED:
       break;
     default:
       break;
   }
 
   return get_status();
+}
+
+void
+ActionExecutor::cancel()
+{
+  state_ = CANCELLED;
+  plansys2_msgs::msg::ActionExecution msg;
+  msg.type = plansys2_msgs::msg::ActionExecution::CANCEL;
+  msg.node_id = current_performer_id_;
+  msg.action = action_name_;
+  msg.arguments = action_params_;
+
+  action_hub_pub_->publish(msg);
 }
 
 std::string
--- a/plansys2_executor/src/plansys2_executor/ActionExecutor.cpp
+++ b/plansys2_executor/src/plansys2_executor/ActionExecutor.cpp
@@ -41,6 +41,8 @@ ActionExecutor::ActionExecutor(
   action_name_ = get_name(action);
   action_params_ = get_params(action);
   completion_ = 0.0;
+  start_execution_ = node_->now();
+  state_time_ = start_execution_;
 }
 
 void
@@ -70,15 +72,20 @@ ActionExecutor::action_hub_callback(const plansys2_msgs::msg::ActionExecution::S
       }
       break;
     case plansys2_msgs::msg::ActionExecution::FEEDBACK:
-      if (state_ != RUNNING || msg->arguments != action_params_ || msg->action != action_name_) {
+      if (state_ != RUNNING || msg->arguments != action_params_ || msg->action != action_name_ ||
+        msg->node_id != current_performer_id_)
+      {
         return;
       }
       feedback_ = msg->status;
       completion_ = msg->completion;
+      state_time_ = node_->now();
 
       break;
     case plansys2_msgs::msg::ActionExecution::FINISH:
-      if (msg->arguments == action_params_ && msg->action == action_name_) {
+      if (msg->arguments == action_params_ &&
+        msg->action == action_name_ && msg->node_id == current_performer_id_)
+      {
         if (msg->success) {
           state_ = SUCCESS;
         } else {
--- a/plansys2_executor/src/plansys2_executor/ActionExecutorClient.cpp
+++ b/plansys2_executor/src/plansys2_executor/ActionExecutorClient.cpp
@@ -84,7 +84,6 @@ CallbackReturnT
 ActionExecutorClient::on_activate(const rclcpp_lifecycle::State & state)
 {
   status_.state = plansys2_msgs::msg::ActionPerformerStatus::RUNNING;
-
   timer_ = create_wall_timer(
     rate_, std::bind(&ActionExecutorClient::do_work, this));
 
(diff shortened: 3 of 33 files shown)


tdenewiler:
I do not see any regressions in my use case when using this branch. It does not fix the issue I am seeing, but I don't see any new issues. For my issue I will work on creating a minimal example and/or getting some gdb output to help resolve. At that time I can make a new issue.

plansys2_bt_actions/src/plansys2_bt_actions/BTAction.cpp
```diff
@@ -53,19 +53,16 @@ BTAction::on_configure(const rclcpp_lifecycle::State & previous_state)
...
   for (auto plugin : plugin_lib_names) {
-    factory.registerFromPlugin(loader.getOSName(plugin));
+    factory_.registerFromPlugin(loader.getOSName(plugin));
   }
 
   auto node = rclcpp::Node::make_shared(std::string(get_name()) + "_aux");
   blackboard_ = BT::Blackboard::create();
   blackboard_->set("node", node);
 
-  tree_ = factory.createTreeFromFile(bt_xml_file_, blackboard_);
```
xydesa, line 67: moving this down to the `on_activate` function will require moving the ZMQ section just below this to the `on_activate` function as well. you may also have to add `publisher_zmq_.reset()` after the call to `haltTree` in `on_deactivate`

plansys2_executor/src/plansys2_executor/ExecutorClient.cpp
```diff
@@ -23,67 +23,146 @@
...
+         status == action_msgs::msg::GoalStatus::STATUS_EXECUTING;
+}
+
+void
+ExecutorClient::cancel_plan_execution()
+{
+  if (should_cancel_goal()) {
+    auto future_cancel = action_client_->async_cancel_goal(goal_handle_);
+    if (rclcpp::spin_until_future_complete(
+        node_->get_node_base_interface(), future_cancel, 3s) !=
```
teyssieuman, line 150: Why are those 3 second for ? Is this possible to have a plan cancellation that last up to 3 seconds ?

teyssieuman: By the way, it is a fantastic work.

plansys2_bt_actions/src/plansys2_bt_actions/BTAction.cpp
xydesa: opened up a PR to add this change to this branch: #93

fmrico: Ok, I will merge after this

plansys2_executor/src/plansys2_executor/ExecutorClient.cpp
fmrico: Well, it's an arbitrary time. It will probably take less, but I have no clear criteria for timeouts until tested in robots with poor connection...

plansys2_executor/src/plansys2_executor/ActionExecutor.cpp
```diff
@@ -207,6 +210,19 @@ ActionExecutor::tick(const rclcpp::Time & now)
   return get_status();
 }
 
+void
+ActionExecutor::cancel()
```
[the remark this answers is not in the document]
fmrico: This message implements the protocol described [here](https://github.com/IntelligentRoboticsLabs/ros2_planning_system/issues/69#issuecomment-761537077). This msg goes to the action performer, so it can't be filled.

plansys2_executor/src/plansys2_executor/ActionExecutorClient.cpp
```diff
@@ -126,6 +125,13 @@ ActionExecutorClient::action_hub_callback(const plansys2_msgs::msg::ActionExecut
         commited_ = false;
       }
       break;
+    case plansys2_msgs::msg::ActionExecution::CANCEL:
+      if (get_current_state().id() == lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE &&
+        msg->node_id == get_name())
+      {
+        trigger_transition(lifecycle_msgs::msg::Transition::TRANSITION_DEACTIVATE);
```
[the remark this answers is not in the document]
fmrico: Yes!! it sounds reasonable. `success` should be `false`, and I could create a member `ActionExecutorClient::completion_` that should be used in the derived classes. What do you think?

plansys2_executor/src/plansys2_executor/ExecutorClient.cpp
teyssieuman: In this case isn't it imaginable to make it a parameter with a default value ?

fmrico:
I will not merge this until we finish the discussion and I finish the experiment with a real robot that I am preparing for an IROS2021 paper. My idea is to test the robustness in a long-duration experiment, similar to what we did in [Marathon2](https://www.youtube.com/watch?v=tXp8wFZr68M).

The idea is to make the real robot to simulate in the laboratory the [factory example](https://github.com/IntelligentRoboticsLabs/ros2_planning_system_examples/tree/master/plansys2_bt_example):
* There are 5 zones: wheels_zone, body_car_zone, steering_wheels_zone, assembly_zone, and recharging_zone.
* The move, transport, and assembly actions are simulated (the robot moves to the areas using Nav2 for transport and move, but instead of assembly, it waits N seconds and considers the action performed)
* The robot needs a predicate that indicates that its battery is full to transport and assembly (overall req)
* The robot recharges in an action that takes N seconds in the recharging_zone
* There is a node that controls the experiment: init and re-init after each assembly the knowledge in problem, request for executing a plan, and when it considers, removes the battery_full predicate, forcing to replan when a requirement fails.

The idea is to run for hours and measure:
* failures
* replans
* actions performed
* total time

Anything that you consider useful to measure easily?

plansys2_executor/src/plansys2_executor/ExecutorClient.cpp
fmrico: Yes, we should make it a parameter in any moment

plansys2_bt_actions/src/plansys2_bt_actions/BTAction.cpp
xydesa: @fmrico that PR is set to merge into this branch, not master. I'd recommend merging it into this branch before merging this branch into master.

plansys2_executor/src/plansys2_executor/ActionExecutor.cpp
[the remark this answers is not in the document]
jjzapf: I think I understand. Actions are handled from 2 sides. The 1st is the performer/client side, which is instantiated by sub-classing the ActionExecutorClient. The 2nd is the executor side, which is an instantiation of ActionExecutor. The two sides communicate via the /action_hub message.

So you're saying that it would not be appropriate to fill in the completion, status, and success fields on the executor side, since those fields are reserved for the performer side, i.e., the performer should tell the executor about its progress not the other way around. Is that correct?

However, would the cancel() function be called from the executor side? In that case, would it be appropriate to inform the performer that the action was canceled on the executor side? One way that I was thinking about using this is to allow the ActionExecutor to cancel the execution if the actual duration of an action exceeded the predicted duration (as specified by the plan) by some percentage. In this scenario, I think cancel() would be invoked from the ActionExecutor.

plansys2_executor/src/plansys2_executor/ActionExecutorClient.cpp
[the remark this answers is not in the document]
jjzapf: Yes. I think that would work.

jjzapf:
> I will not merge this until we finish the discussion and I finish the experiment with a real robot that I am preparing for an IROS2021 paper. My idea is to test the robustness in a long-duration experiment, similar to what we did in [Marathon2](https://www.youtube.com/watch?v=tXp8wFZr68M).
> 
> The idea is to make the real robot to simulate in the laboratory the [factory example](https://github.com/IntelligentRoboticsLabs/ros2_planning_system_examples/tree/master/plansys2_bt_example):
> 
> * There are 5 zones: wheels_zone, body_car_zone, steering_wheels_zone, assembly_zone, and recharging_zone.
> * The move, transport, and assembly actions are simulated (the robot moves to the areas using Nav2 for transport and move, but instead of assembly, it waits N seconds and considers the action performed)
> * The robot needs a predicate that indicates that its battery is full to transport and assembly (overall req)
> * The robot recharges in an action that takes N seconds in the recharging_zone
> * There is a node that controls the experiment: init and re-init after each assembly the knowledge in problem, request for executing a plan, and when it considers, removes the battery_full predicate, forcing to replan when a requirement fails.
> 
> The idea is to run for hours and measure:
> 
> * failures
> * replans
> * actions performed
> * total time
> 
> Anything that you consider useful to measure easily?

This sounds really cool!! Just a couple of thoughts for you to consider.

1. For our work, we have simulated the robot battery using a Gazebo plugin. @xydesa did this work and could provide more information if you're interested. We then used a function on the PDDL side to predict the state of charge of the robot over it's mission rather than using a predicate. You can find a simpler version of this here (https://github.com/IntelligentRoboticsLabs/ros2_planning_system/blob/master/plansys2_executor/test/pddl/domain_charging.pddl#L29). In our internal repos we have an improved version, which does a better job of tracking the Gazebo simulated battery.
2. I would be interested to hear what your thoughts are regarding re-plan triggers. This is something that we have been considering, particularly as it relates to long term missions. Perhaps we could start another discussion somewhere to discuss this in more detail.

plansys2_executor/src/plansys2_executor/ExecutorClient.cpp
```diff
@@ -23,67 +23,146 @@
...
+bool
+ExecutorClient::execute_and_check_plan()
+{
+  if (rclcpp::ok() && !goal_result_available_) {
+    rclcpp::spin_some(node_);
 
-  if (!this->execute_plan_client_ptr_) {
-    RCLCPP_ERROR(node_->get_logger(), "Action client not initialized");
+    if (!goal_result_available_) {
+      return true;  // Plan not finished
```
jjzapf, line 72: This seems to contradict your description in the PR discussion. ...

"This method must be called iteratively to refresh the result and feedback. It returns true if the plan execution finishes."

Should this return be false and the other true?

fmrico: Yes, you are right. I noticed this yesterday when preparing the experiment. It returns true while the plan is still executing.

I will edit the PR description unless you think it's more appropriate to return true when finishes.

plansys2_bt_actions/src/plansys2_bt_actions/BTAction.cpp
fmrico: Merged!! :wink:

plansys2_executor/src/plansys2_executor/ActionExecutor.cpp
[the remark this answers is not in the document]
fmrico: > So you're saying that it would not be appropriate to fill in the completion, status, and success fields on the executor side, since those fields are reserved for the performer side, i.e., the performer should tell the executor about its progress not the other way around. Is that correct?

Exactly

> However, would the cancel() function be called from the executor side? In that case, would it be appropriate to inform the performer that the action was canceled on the executor side? 

We do it. An ActionExecution.msg is sent from the Executor side, setting the type field to CANCEL. These are the possible values for type (from ActionExecution.msg):

`
int16 REQUEST=1
int16 RESPONSE=2
int16 CONFIRM=3
int16 REJECT=4
int16 FEEDBACK=5
int16 FINISH=6
int16 CANCEL=7
`

> One way that I was thinking about using this is to allow the ActionExecutor to cancel the execution if the actual duration of an action exceeded the predicted duration (as specified by the plan) by some percentage. In this scenario, I think cancel() would be invoked from the ActionExecutor.

This is interesting. But I am not sure if using the duration specified in the domain file is adequate. The duration of some actions is impossible to really predict (navigation, for example). Maybe we could add an optional parameter in the ActionExecutionClient to set a timeout...

fmrico:
> 1. For our work, we have simulated the robot battery using a Gazebo plugin. @xydesa did this work and could provide more information if you're interested. We then used a function on the PDDL side to predict the state of charge of the robot over it's mission rather than using a predicate. You can find a simpler version of this here (https://github.com/IntelligentRoboticsLabs/ros2_planning_system/blob/master/plansys2_executor/test/pddl/domain_charging.pddl#L29). In our internal repos we have an improved version, which does a better job of tracking the Gazebo simulated battery.

It sounds quite interesting!! I'll love to know more about it

> 2. I would be interested to hear what your thoughts are regarding re-plan triggers. This is something that we have been considering, particularly as it relates to long term missions. Perhaps we could start another discussion somewhere to discuss this in more detail.

Yes, we have to open an issue for this. 

I always considered replanning as "execute again the plan", but maybe there are more factors to take into account.

I am also concerned about the cancelation (via cancel or by over all req) of actions that removes predicates at_start and add predicates at_end. For example, the move action in many examples of the repo, that removes robot_at at initial, and sets another robot_at predicate at_end. If you cancel it, you lost the knowledge that indicates where the robot is.

jjzapf:
> It sounds quite interesting!! I'll love to know more about it

I think we should be able release this. I will need to talk with @xydesa about it.

> I am also concerned about the cancelation (via cancel or by over all req) of actions that removes predicates at_start and add predicates at_end. For example, the move action in many examples of the repo, that removes robot_at at initial, and sets another robot_at predicate at_end. If you cancel it, you lost the knowledge that indicates where the robot is.

Yes, I ran into this problem with some experiments that I was doing. I'm not sure if this is the best way to handle it, but here is what I did. In the controller, if the execution fails or is cancelled I first retrieve all the predicates via a call to getPredicates. I then search for any predicate in the set of the form "roboat_at". If there are no such predicates, then the robot location is undefined (as far as the planner side is concerned). I then add a new predicate via the addPredicate function of the form (robot_at r2d2 wp_current). The waypoint value associated with wp_current is updated regularly in the current_pos_callback function in the MoveAction. Thus when the planner restarts (i.e. computes a new plan) the robot location is defined.

plansys2_executor/src/plansys2_executor/ExecutorClient.cpp
jjzapf: No. I think it's fine the way you have it. Returning true to indicate that it's still executing makes sense to me. Just double checking that this was the intended behavior.

fmrico:
The ongoing experiment https://youtu.be/2yCuZLhCzKk

fmrico:
After finalizing the experiments, fixing bugs, and making some improvements, I am merging this.

PR #96, opened 2021-03-04
Action result tweaks

kogut:
Signed-off-by: Greg Kogut <kogut@spawar.navy.mil>

This corrects two edge cases.  One is where a BehaviorTree action return `NodeStatus::FAILURE` but does not appropriately return execution feedback status indicating failure. This can result in a plan considering the action a success and moving on to the next action.   The change assumes that a failed BT is a failed action.

Another change corrected an error where the completion percentage would be set to 1.0 on the completion of action regardless of what the actual completion percent was.    E.g. just because an action has finished probably shouldn't imply that 100% of the requested action was accomplished.  

It was not trivial to provide test instructions to test these edge cases using the usual example scenario.  However running the plansys2 tutorial example should indicate no regressions.

### DIFF
--- a/plansys2_executor/src/plansys2_executor/ExecutorNode.cpp
+++ b/plansys2_executor/src/plansys2_executor/ExecutorNode.cpp
@@ -274,10 +274,12 @@ ExecutorNode::execute(const std::shared_ptr<GoalHandleExecutePlan> goal_handle)
   if (status == BT::NodeStatus::FAILURE) {
     tree.haltTree();
     RCLCPP_ERROR(get_logger(), "Executor BT finished with FAILURE state");
+    result->success = false;
+  } else {
+    result->success = true;
   }
 
   result->action_execution_status = get_feedback_info(action_map);
-  result->success = true;
 
   size_t i = 0;
   while (i < result->action_execution_status.size() && result->success) {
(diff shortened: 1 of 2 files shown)


plansys2_executor/src/plansys2_executor/ExecutorNode.cpp
```diff
@@ -274,10 +274,12 @@ ExecutorNode::execute(const std::shared_ptr<GoalHandleExecutePlan> goal_handle)
   if (status == BT::NodeStatus::FAILURE) {
     tree.haltTree();
     RCLCPP_ERROR(get_logger(), "Executor BT finished with FAILURE state");
+    result->success = false;
+  } else {
+    result->success = true;
   }
 
   result->action_execution_status = get_feedback_info(action_map);
-  result->success = true;
```
fmrico, line 280: maybe we could write?

```
result->success = status == BT::NodeStatus::SUCCESS;
```

fmrico (changes requested):

plansys2_executor/src/plansys2_executor/ExecutorNode.cpp
kogut: Of course.  b522c0b

kogut: Of course.   56ef8b6729fb7f3d9ae551c0d23cb4e17b522c0b

fmrico (approved):

PR #100, opened 2021-03-10
Fix typo in comment string to stop LCOV exclusion block.

tdenewiler:
Signed-off-by: Thomas Denewiler <thomas.denewiler@navy.mil>


fmrico:
Good catch 

Thanks!!

PR #101, opened 2021-03-11
Add check for empty pointer prior to accessing pointer. Applies to ch…

tdenewiler:
…ecking the tree types of a predicate, involved in checking if a goal is valid.

Signed-off-by: Thomas Denewiler <thomas.denewiler@navy.mil>


fmrico:
Merging. Thanks!! :)

PR #103, opened 2021-03-16
Get ordered sub goals

xydesa:
Add GetOrderedSubGoals service to Executor, allowing executor clients to get the order in which sub-goals will be completed by the current plan.

This will allow for things like removing a specific sub-goal if the system keeps failing to achieve that sub-goal.


fmrico:
LGTM, Thanks!!!

PR #109, opened 2021-04-01
Reducing log message severity because lack of a plan isn't necessaril…

kogut:
…y an error.

Signed-off-by: Greg Kogut <kogut@spawar.navy.mil>


fmrico:
Thanks @kogut 

It sounds reasonable and nobody disagreed with it these days. So, merging...

PR #110, opened 2021-04-02
Action execution timeout

jjzapf:
In some applications it may be important that the actual action execution duration closely match the predicted action execution duration defined by the plan. In these cases, a duration overrun percentage can be passed to the system which defines an allowable time overrun of an action based on a percentage of the predicted plan duration. For example, if the plan predicts that an action should take 1000 secs and a duration overrun percentage of 20% is specified, then the action should be halted if the actual duration exceeds 1200 secs.

This PR will add an action timeout feature to PlansSys2. A duration_overrun_percentage parameter has been added to the ExecutorNode. Likewise, a string array performer_parameters parameter has been added to the ActionExecutorClient, from which a complementary duration_overrun_percentage may be extracted. In each case, the percentage value defaults to -1. Only non-negative values will engage the action timeout capability. If only the ExecutorNode percentage value is set to be non-negative, then all actions will halt according to the time specified by their durations and the "global" duration_overrun_percentage. If only the ActionExecutorClient percentage values are set, then only the associated actions will obey the timeout constraints, as specified per performer. If the ExecutoNode percentage value is set and one or more of the ActionExecutorClient percentage values are set, then each non-negative ActionExecutorClient percentage will override the "global" value set by the ExecutorNode.

Two new unit tests have been added to verify this capability.


fmrico:
Hi @jjzapf 

I see a lot of changes, even in the protocol, and I am not sure if making it more complex is a good idea. I think that this deserves a more in-deep discussion. 

I see this feature as an action requirement. Maybe it is cleaner to add this feature as a new BT action node in `plansys2_executor`. Now each action is expanded to an "action BT"  like:

```
              <Sequence name="(move leia dinning bedroom):15004">
                <WaitAtStartReq action="(move leia kitchen dinning):10003"/>
                <ApplyAtStartEffect action="(move leia dinning bedroom):15004"/>
                <ReactiveSequence name="(move leia dinning bedroom):15004">
                  <CheckOverAllReq action="(move leia dinning bedroom):15004"/>
                  <ExecuteAction action="(move leia dinning bedroom):15004"/>
                </ReactiveSequence>
                <CheckAtEndReq action="(move leia dinning bedroom):15004"/>
                <ApplyAtEndEffect action="(move leia dinning bedroom):15004"/>
              </Sequence>
```

This is now hardcoded, but maybe we could have it in a separate file and used when generating the execution Behavior Tree. The user could adapt this action BT to its needs. You could have something like:

```
              <Sequence name="(move leia dinning bedroom):15004">
                <WaitAtStartReq action="(move leia kitchen dinning):10003"/>
                <ApplyAtStartEffect action="(move leia dinning bedroom):15004"/>
                <ReactiveSequence name="(move leia dinning bedroom):15004">
                  <CheckDuration action="(move leia dinning bedroom):15004" tolerance="0.20" />             
                  <CheckOverAllReq action="(move leia dinning bedroom):15004"/>
                  <ExecuteAction action="(move leia dinning bedroom):15004"/>
                </ReactiveSequence>
                <CheckAtEndReq action="(move leia dinning bedroom):15004"/>
                <ApplyAtEndEffect action="(move leia dinning bedroom):15004"/>
              </Sequence>
```

This node is checked in every cycle and could return FAILURE if the time is exceeded. Then the plan fails, all the actions are halted, and it replans. 

What do you think?

jjzapf:
Hi @fmrico

Thanks for looking at these changes. I'm definitely open to other ideas on how to implement this. I think another BT action node, i.e., a CheckDuration node, could work. I'm currently using the BehaviorTree.CPP timeout feature to do this. This built in feature seems appropriate for constraining the amount of time that an action has to perform it's task, but I could try the CheckDuration node approach if you think that is better.

I'm not sure I understand your suggestion about putting the "action BT" in a separate file. The XML code is currently auto generated in the execution_block function of BTBuilder.cpp. It seems like you are suggesting that we no longer auto generate this code, but leave it for the user to pass in as a file. I'm not sure how this would work and it seems like it would be much more prone to user error. Could you elaborate on this idea a little more? Currently, i'm using the timeout arg that is passed into the execution_block function to determine whether or not the timeout portion of the XML code should be generated or not. The timeout XML code is only generated when the value is greater than 0. By default it is set to -1.

I believe the protocol changes you mentioned refer to the SETUP_REQUEST and SETUP_RESPONSE message types that I added to the ActionExecution message. These were added because I needed a way to pass the duration_overrun_percentage parameter from an action executor client node to it's associated action executor. Without this communication, we could not set individual values of the duration_overrun_percentage for each action. This feature is important, because we do not want to treat each action the same. That is, for some actions we would like to use the timeout feature and for others we do not.

One possibility for reducing the complexity of the protocol between the action executor client and it's action executor might be to use the response callback mechanism of ROS2 services. This is something that I was unfamiliar with until @xydesa showed it to me recently. Basically, you can specify a response callback when a request is made by the service client, which seems to be what is needed for communicating between the action executor client and it's action executor. We could look into this further if you think it might be a good idea.

fmrico:
Hi @jjzapf 


> I'm not sure I understand your suggestion about putting the "action BT" in a separate file. The XML code is currently auto generated in the execution_block function of BTBuilder.cpp. It seems like you are suggesting that we no longer auto generate this code, but leave it for the user to pass in as a file. I'm not sure how this would work and it seems like it would be much more prone to user error. Could you elaborate on this idea a little more? 

Check #111, please. 

> Currently, i'm using the timeout arg that is passed into the execution_block function to determine whether or not the timeout portion of the XML code should be generated or not. The timeout XML code is only generated when the value is greater than 0. By default it is set to -1.

The problem is starting to add default parameters to general functions that are useful for a specific feature. I am concerned about starting to add complex or not obvious functions and parameters. This could be an obstacle in the life of PlanSys2.

> I believe the protocol changes you mentioned refer to the SETUP_REQUEST and SETUP_RESPONSE message types that I added to the ActionExecution message. These were added because I needed a way to pass the duration_overrun_percentage parameter from an action executor client node to it's associated action executor. Without this communication, we could not set individual values of the duration_overrun_percentage for each action. This feature is important, because we do not want to treat each action the same. That is, for some actions we would like to use the timeout feature and for others we do not.

Yes, #111 does not support individual timeouts. Let me think about it.

> One possibility for reducing the complexity of the protocol between the action executor client and it's action executor might be to use the response callback mechanism of ROS2 services. This is something that I was unfamiliar with until @xydesa showed it to me recently. Basically, you can specify a response callback when a request is made by the service client, which seems to be what is needed for communicating between the action executor client and it's action executor. We could look into this further if you think it might be a good idea.

Actually, the good mechanism would be a ROS2 action, which was the original implementation. The auction protocol was made for improving this mechanism...

fmrico:
For individual timeouts, we need to add something like this in `plansys2_params.yaml`:

```
planner:
  ros__parameters:
    plan_solver_plugins: ["POPF"]
    POPF:
      plugin: "plansys2/POPFPlanSolver"
    TFD:
      plugin: "plansys2::TFDPlanSolver"

executor:
  ros__parameters:
    action_timeouts: ["move", "assemble"]
    move: 0.20
    assemble: 0.45
```

This value should go directly to the [`action_map` ](https://github.com/IntelligentRoboticsLabs/ros2_planning_system/blob/47c4e6f1cccf6e6a5683a75c44ba217f73c8f55a/plansys2_executor/src/plansys2_executor/ExecutorNode.cpp#L273) by adding extra field to `ActionExecutionInfo`. Then, it could be accessed by any custom BT node, like [here](https://github.com/IntelligentRoboticsLabs/ros2_planning_system/blob/47c4e6f1cccf6e6a5683a75c44ba217f73c8f55a/plansys2_executor/src/plansys2_executor/behavior_tree/wait_action_node.cpp#L40).

I am trying to think of an alternative to avoid modifying `ActionExecutionInfo`, but I have not found a solution. What is certain that I want to avoid at all costs is to modify the protocol and that the responsibility for canceling an action lies with the ActionExecutorClient.

xydesa:
@fmrico I've updated this PR to address your concerns: I've removed all the changes to the actions_hub protocol between ActionExecutor and ActionExecutorClient. And I've added node parameters for specifying individual action timeouts on the ExecutorNode (and included README changes specifying how they work). It also handles the case where those parameters are not declared (and therefore no timeouts needed).

It still uses the `Timeout` node built into BehaviorTree.CPP to handle the timeouts.
I also left the action duration field on the ActionExecutionInfo message because I felt like this could be useful to a developer interacting with the ExecutorNode via the ExecutorClient.

jjzapf:
Thanks for making these changes @xydesa. And thanks for the original suggestion @fmrico. I think it was a good call to remove the changes that I originally made to the action_hub protocol.

fmrico:
Hi @jjzapf 

First of all, I have merged #111 because I think that this will be useful in this PR.

I have analyzed the code, and I have a suggestion. As you know, I prefer don't change `BTBuilder` to include duration or the action map in the headers. I would like to propose a change that could avoid this and make it simpler to achieve your objectives.

I have seen that the `ExecutorNode` reads the parameters related to the specific percentage of each action. Good. The `ExecutorNode` also has the expected action duration from the planner. The `ActionExecutionInfo`s in the `action_map`  has a `duration` and a `duration_overrun_percentage` field. Good. So, we have all we need. We don't need to change BTbuilder. 

Let's consider that your custom `plansys2_action_bt.xml` is:

```
<Sequence name="ACTION_ID">
WAIT_AT_START_ACTIONS
    <CustomTimeout action="ACTION_ID"/>
  </ReactiveSequence>
</Sequence>
```

The implementation of `CustomTimeout` could directly access the `action_map` to check all it needs to check the deadline. Something like:

``` 
BT::NodeStatus
CustomTimeout::tick()
{
  std::string action;
  getInput("action", action);

 if (status() == BT::NodeStatus::IDLE) {
    start_ = std::chrono::high_resolution_clock::now();
  } 

  if ((*action_map_)[action].action_executor != nullptr)
  {
    float max_duration =  (*action_map_)[action].durative_action_info->duration_overrun_percentage * 
         (*action_map_)[action].durative_action_info->duration; 
    
   
   if (std::chrono::high_resolution_clock::now() - start_ >  max_duration) {
      return BT::NodeStatus::FAILURE;
   } else {    
      return BT::NodeStatus::SUCCESS;
  }
}
```

I think that this is all you need. You wouldn't need to change `BTBuilder`. What do you think?.

fmrico:
Thanks, @jjzapf :blush:

jjzapf:
Hi @fmrico

I've updated this PR with the changes that we discussed above. We're still reviewing this internally, but I wanted to give you a chance to look at it now. I don't imagine there will be many changes requested from my internal reviewers. If you could hold off on merging this until they're done I'd appreciate it. Please feel free to start commenting on this now, however, with any suggestions you may have. Thanks!

fmrico:
Hi @jjzapf 

 I have seen you closed the PR. Are you preparing another? I had no time to have a look at it, sorry :S

jjzapf:
Hi @fmrico 

Yes. I opened a new one. I wanted to clean up the history, since I accidentally included a lot of our internal commits on the old PR. I wasn't sure the best way to do this. I ended up just deleting the old branch and applying the diff to a new branch created from master.

PR #111, opened 2021-04-11
Configurable Behavior Tree for actions

fmrico:
Dear PlanSys2 developers

When building the BT that codes the plan, the generation of the last part, corresponding to the actions, is hardcoded. 

In #110, @jjzapf proposed a PR related to add QoS functionality to the actions, and we started to discuss that maybe a clean solution is set up this action BT. This PR is the first step for this solution.

In this PR, we can configure the action BT in a separate file. By default, the action BT is in the `behavior_trees` folder in `plansys2_executor`, and contains:

```
<Sequence name="ACTION_ID">
WAIT_AT_START_ACTIONS
  </ReactiveSequence>
</Sequence>
```

Now you can select an alternative action BT, adding custom BT nodes that you can implement in the same way the other BT nodes (`ApplyAtStartEffect`, `CheckAtEndReq` and so on). 

It is important to respect some rules in your custom action BT:
1. It should be a sequence `<Sequence name="ACTION_ID">`
2. Every BT node has an `action="ACTION_ID"/` property to refer to the internal structures correctly.
3. `WAIT_AT_START_ACTIONS` must be the first element to guarantee the correct execution of the tree.

I hope it helps!!

### DIFF
--- a/plansys2_executor/src/plansys2_executor/BTBuilder.cpp
+++ b/plansys2_executor/src/plansys2_executor/BTBuilder.cpp
@@ -38,6 +38,9 @@ BTBuilder::BTBuilder(
 {
   domain_client_ = std::make_shared<plansys2::DomainExpertClient>(node);
   problem_client_ = std::make_shared<plansys2::ProblemExpertClient>(node);
+
+  std::string action_bt_xml_filename =
+    get_parameter("default_action_bt_xml_filename").get_value<std::string>();
 }
 
 void
@@ -500,6 +503,10 @@ BTBuilder::get_graph(const Plan & current_plan)
     remove_existing_requirements(at_start_requirements, predicates, functions);
     remove_existing_requirements(over_all_requirements, predicates, functions);
     remove_existing_requirements(at_end_requirements, predicates, functions);
+
+    for (const auto & req : at_start_requirements) {
+      std::cerr << "===> [" << req->toString() << "]" << std::endl;
+    }
 
     assert(at_start_requirements.empty());
     assert(over_all_requirements.empty());
--- a/plansys2_executor/src/plansys2_executor/BTBuilder.cpp
+++ b/plansys2_executor/src/plansys2_executor/BTBuilder.cpp
@@ -43,7 +43,8 @@ BTBuilder::BTBuilder(
   if (bt_action != "") {
     bt_action_ = bt_action;
   } else {
-    bt_action_ = R""""(<Sequence name="ACTION_ID">
+    bt_action_ =
+      R""""(<Sequence name="ACTION_ID">
 WAIT_AT_START_ACTIONS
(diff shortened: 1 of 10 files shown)


plansys2_executor/src/plansys2_executor/BTBuilder.cpp
```diff
@@ -709,31 +731,46 @@ BTBuilder::t(int level)
...
 
-  ret = ret + t(l + 1) + "<ApplyAtStartEffect action=\"" + action_id + "\"/>\n";
-  ret = ret + t(l + 1) + "<ReactiveSequence name=\"" + action_id + "\">\n";
-  ret = ret + t(l + 2) + "<CheckOverAllReq action=\"" + action_id + "\"/>\n";
-  ret = ret + t(l + 2) + "<ExecuteAction action=\"" + action_id + "\"/>\n";
-  ret = ret + t(l + 1) + "</ReactiveSequence>\n";
-  ret = ret + t(l + 1) + "<CheckAtEndReq action=\"" + action_id + "\"/>\n";
-  ret = ret + t(l + 1) + "<ApplyAtEndEffect action=\"" + action_id + "\"/>\n";
-  ret = ret + t(l) + "</Sequence>\n";
+  replace(ret_aux, "ACTION_ID", action_id);
```
xydesa, line 764: couldn't you just use `std::replace` from `<algorithm>` here instead of writing your own function?
https://stackoverflow.com/questions/2896600/how-to-replace-all-occurrences-of-a-character-in-string
https://www.cplusplus.com/reference/algorithm/replace/

plansys2_executor/src/plansys2_executor/BTBuilder.cpp
```diff
@@ -709,31 +731,46 @@ BTBuilder::t(int level)
...
+  replace(ret_aux, "WAIT_AT_START_ACTIONS", wait_actions);
```
xydesa, line 765: I think this instance should also use either the `std::replace` or just the regular `string.replace` since (if i understand everything correctly) there's only a single instance of `WAIT_AT_START_ACTIONS` that you want to relace.

plansys2_executor/src/plansys2_executor/BTBuilder.cpp
fmrico: Absolutely!!

plansys2_executor/src/plansys2_executor/BTBuilder.cpp
fmrico: ok

plansys2_executor/src/plansys2_executor/BTBuilder.cpp
fmrico: Hi @xydesa 

I have just checked that this function only replaces one char :(

PR #116, opened 2021-04-16
Making TFD plugin type match popf.

xydesa:
Signed-off-by: Alexander Xydes <alexander.xydes@navy.mil>

Based on this commit: https://github.com/IntelligentRoboticsLabs/ros2_planning_system/commit/f5478d3370ce4f54b6e7f2a4c9fdffac6180461f

and testing by switching the popf type to use `::` (which failed). I think the TFD plugin type should use a slash.

### DIFF
--- a/plansys2_bringup/params/plansys2_params.yaml
+++ b/plansys2_bringup/params/plansys2_params.yaml
@@ -4,4 +4,4 @@ planner:
     POPF:
       plugin: "plansys2/POPFPlanSolver"
     TFD:
-      plugin: "plansys2::TFDPlanSolver"
+      plugin: "plansys2/TFDPlanSolver"


fmrico:
Well, I think that it depends on what is the tag `name` in `plansys2_tfd_plan_solver_plugin.xml`, what make me notice that this is missing in this file. It should be:

```
<class_libraries>
  <library path="plansys2_tfd_plan_solver">
    <class name="plansys2/TFDPlanSolver" type="plansys2::TFDPlanSolver" base_class_type="plansys2::PlanSolverBase">
      <description></description>
    </class>
  </library>
</class_libraries>
```

I will change right now, and accept this.

Thanks!!

PR #119, opened 2021-04-27
Action execution timeout

jjzapf:
In some applications it may be important that the actual action execution duration closely match the predicted action execution duration defined by the plan. In these cases, a duration overrun percentage can be passed to the system which defines an allowable time overrun of an action based on a percentage of the predicted plan duration. For example, if the plan predicts that an action should take 1000 secs and a duration overrun percentage of 20% is specified, than the action should be halted if the actual duration exceeds 1200 secs.


jjzapf:
Hi @fmrico

I realized that I accidentally pushed a lot of our internal commits to the original PR for this change. I therefore deleted that branch and pushed a clean branch with a single commit. Hence the new PR. I have screen shots of our original conversation. I will add that conversation below so that it is not lost.

jjzapf:
From @fmrico ...

Hi @jjzapf 

First of all, I have merged #111 because I think that this will be useful in this PR.

I have analyzed the code, and I have a suggestion. As you know, I prefer don't change BTBuilder to include duration or the action map in the headers. I would like to propose a change that could avoid this and make it simpler to achieve your objectives.

I have seen that the ExecutorNode reads the parameters related to the specific percentage of each action. Good. The ExecutorNode also has the expected action duration from the planner. The ActionExecutionInfo in the action_map has duration and a duration_overrun_percentage field. Good. So, we have all we need. We don't need to change BTBuilder.

Let's considr that your custom plansys2_action_bt.xml is:

```
<Sequence name="ACTION_ID">
WAIT_AT_START_ACTIONS
    <CustomTimeout action="ACTION_ID"/>
  </ReactiveSequence>
</Sequence>
```
The implementation of CustomTimeout could directly access the action_map to check all it needs to check the deadline. Something like:

```
CheckTimeout::CheckTimeout(
  const std::string & xml_tag_name,
  const BT::NodeConfiguration & conf)
: ActionNodeBase(xml_tag_name, conf)
{
  action_map_ =
    config().blackboard->get<std::shared_ptr<std::map<std::string, ActionExecutionInfo>>>(
    "action_map");
}

BT::NodeStatus
CheckTimeout::tick()
{
  std::string action;
  getInput("action", action);

  if (status() == BT::NodeStatus::IDLE) {
    start_ = std::chrono::high_resolution_clock::now();
  }

  if ((*action_map_)[action].action_executor != nullptr) {
    float
    double duration = (*action_map_)[action].duration;
    double duration_overrun_percentage = (*action_map_)[action].duration_overrun_percentage;
    if (duration_overrun_percentage >= 0) {
      double max_duration = (1.0 + duration_overrun_percentage / 100.0) * duration;
      auto current_time = std::chrono::high_resolution_clock::now();
      auto elapsed_time = std::chrono::duration_cast<std::chrono::milliseconds>(
        current_time - start_);
      if (elapsed_time > std::chrono::duration<double>(max_duration)) {
        std::cerr << "Actual duration of " << action << " exceeds max duration (" << std::fixed <<
          std::setprecision(2) << max_duration << " secs)." << std::endl;
        return BT::NodeStatus::FAILURE;
      }
    }
  }

  return BT::NodeStatus::SUCCESS;
}

File: plansys2_executor/test/test_behavior_trees/plansys2_action_bt.xml [hunk 1]
```diff
@@ -0,0 +1,11 @@
+<Sequence name="ACTION_ID">
+WAIT_AT_START_ACTIONS
+  <ApplyAtStartEffect action="ACTION_ID"/>
+  <ReactiveSequence name="ACTION_ID">
+    <CheckTimeout action="ACTION_ID"/>
```
fmrico, line 5: I would leave the default BT without the `CheckTimeout` node because it is particular to an application that needs it. In case of needing it, you can pass your custom action BT with the `default_action_bt_xml_filename` parameter.

fmrico (changes requested):

fmrico:
Hi @jjzapf 

This PR is just what we talked about, so I will be very happy to accept it as is, only this the change I propose to you in the review.

Thanks!!!

fmrico (approved):

plansys2_executor/test/test_behavior_trees/plansys2_action_bt.xml
jjzapf: I haven't actually changed the default BT. I just created a test_behavior_tree folder in the test directory and added this BT to test the action timeout capability. See action_timeout test in executor_test.cpp. To make this a bit clearer, however, I changed the name of this file from plansys2_action_bt.xml to test_action_timeout_bt.xml.

fmrico:
Merged!

Thanks, @jjzapf !!

PR #134, opened 2021-06-14
Compile for ROS2 Galactic

jonatanolofsson:
This PR fixes compilation for galactic, but breaks it for foxy. A new branch should be created before its merge.


fmrico:
Hi @jonatanolofsson 

Thanks for your contribution.

I have created a branch for Galactic, and I have edited this PR to point there. In a couple of weeks, I plan to release the package for Foxy and change the `master` branch to galactic. Works for Foxy (only bug fixes) will be in `foxy-devel`.

Maybe I have to adjust CI for this PR. Let's see...

PR #139, opened 2021-06-14
Pddl domain (:constants ) handling and planning

MostafaGomaa:
This PR fixes handling of pddl domain constants:

Fixes:
[domain_expert] : parsing "(:constants)" from domain_string
[pddl_parser] : Ground::getTree correctly grounds constants values when grounding actions conditions, as well as, when adding predicates
[problem_expert] : 
     - add domain constants as `problem_expert::instances` when adding a problem.pddl from file
     - when generating a problem.pddl file (via `getProblem()`, exclude constants from "(:objects)" . Otherwise, popf will print warnings and failed to generate a plan
```
problem.pddl: line: 4: Warning: Re-declaration of symbol in same scope: jack
problem.pddl: line: 6: Warning: Re-declaration of symbol in same scope: leia

The planner will continue, but you may wish to fix your files accordingly
Constructing lookup tables: [10%] [20%] [30%] [40%] [50%] [60%] [70%] [80%] [90%] [100%] [110%] [120%]
A problem has been encountered, and the problem has been deemed unsolvable
--------------------------------------------------------------------------
The goal fact:
(robot_talk leia m1 jack)

...cannot be found either in the initial state, as an add effect of an
 action, or as a timed initial literal.  As such, the problem has been deemed
unsolvable.
```

Features:
*  [domain_expert] & [msgs] GetDomainConstants service 
* Unit-tests for changes

Minor Fixes unrelated to the constants but caused test failure:
* [problem_expert] also clearing goals upon clearance of knowledge; this caused silent termination of problem_expert node when `getProblem()` was called after a call to `clearKnowledge()`.

Fix #138 , #135

PR #160, opened 2021-09-14
Minor fixes

MostafaGomaa:
This PR fixes minor issues with 
* [domain_expert] styling divergence in DomainExpertInterface; this caused errors in running the tests
* [problem_expert] also clearing goals upon clearance of knowledge; this caused silent termination of problem_expert node when `getProblem()` was called after a call to `clearKnowledge()`.

### DIFF
--- a/plansys2_problem_expert/src/plansys2_problem_expert/ProblemExpert.cpp
+++ b/plansys2_problem_expert/src/plansys2_problem_expert/ProblemExpert.cpp
@@ -382,6 +382,8 @@ ProblemExpert::clearKnowledge()
   instances_.clear();
   predicates_.clear();
   functions_.clear();
+  clearGoal();
+
   return true;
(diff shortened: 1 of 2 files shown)


MostafaGomaa:
hmm..strange, the cli fails here (even though it works on mine) for the opposite reason if fails on my machine.

Screenshot_2021-09-16_15-05-29


I.e., My commit breaks it here but fixes it on my machine!

Could it be different versions of uncrustify?

anyway, I reverted the problematic change.

MostafaGomaa:
The changes are now included in #139

fmrico:
Hi @MostafaGomaa 

Sometimes I have suffered the same situation, and with things like these spaces: in local, everything runs ok, but CI fails. Usually, GitHub is the reference.

Thanks!!

PR #163, opened 2021-11-11
Adding service to retrieve domain name.

jjzapf:
In our client software we have specialized code that depends on the domain that is loaded. I've therefore added a service to the domain expert to get the domain name. I trust this will be useful to other users as well.

I've also modified the way that the domain name is generated when combining domains. The previous method was to use "plansys2" for the combined domain name. I've modified this so that the new domain name is a concatenation of the individual domain names separated by underscores.

Finally, I added a unit test to verify the new service.


fmrico:
Thanks @jjzapf !!!

Issue #175, opened 2022-02-02
Unnecessary node pointer in `execute_action_node.cpp`


sarcasticnature:
The node pointer being retrieved from the blackboard in the line below is only used to get a logger for the `RCLCPP_ERROR_STREAM` statement on line 70 (of the same file).  Given that there is a `node_` member variable of the class, I'm assuming this is a typo

https://github.com/IntelligentRoboticsLabs/ros2_planning_system/blob/9247f610916996dbec78d1e0b5ec944f38486905/plansys2_executor/src/plansys2_executor/behavior_tree/execute_action_node.cpp#L55

sarcasticnature:
Planning to fix in my upcoming PR to solve #174, unless this is intentional

Issue #174, opened 2022-02-02
BTActionNode should return BT::NodeStatus::FAILURE instead of throwing std::runtime_error


sarcasticnature:
Currently the `BTActionNode` class will throw a runtime error if sending a goal to the ROS2 action server fails or the server rejects the goal.  This should be changed to returning a `BT::NodeStatus::FAILURE` so that the behavior tree can handle the error with recovery behavior(s), instead of simply dying.  If recovery behaviors are not implemented/fail, the `ActionExecutor` will properly register that the action has failed instead of waiting indefinitely for a message on `/actions_hub` that will never come.

Also `BTActionNode` currently does not provided a means for a child class to indicate a failure during execution (only throw a runtime/logic/etc. error in the `on_tick` method), creating the same issue as above.  This should also be changed to allow a `BT::NodeStatus::FAILURE` when necessary.

sarcasticnature:
I'm currently working on this issue, but wanted to file this issue to let others know and get any feedback @fmrico might have

fmrico:
Hi @sarcasticnature 

You are right!! I think this is a good improvement. Go ahead, please.

Thanks!!

sarcasticnature:
I'll also be implementing the (optional) port to specify a server timeout, it looks like it was started but never finished.

Issue #161, opened 2021-09-14
Throw useful exceptions instead of failing/terminating silently


MostafaGomaa:
This is to document places in the code where the program terminates silently instead of throwing a useful exception. This makes it hard to understand what actually happened or to recover in run-time.

This collaborative issue is intended to spark a discussion about:
* Whether this really is an intended behavior
* What's the best way to notify the user
* What's the best way to recover from such run-time issue

Please, feel free to mark code locations where this happens, or to comment!

MostafaGomaa:
https://github.com/IntelligentRoboticsLabs/ros2_planning_system/blob/e713d72a4aa617d5e50439b8afe079d84851e8f2/plansys2_pddl_parser/src/plansys2_pddl_parser/TypeGround.cpp#L28

fmrico:
Hi @MostafaGomaa 

This is clearly a bug or a design error. We should address this soon.

Thanks

sarcasticnature:
I'm currently working on adding/fleshing out failure messages to some stuff, as part of a PR to fix #174.  So far I have implemented a more descriptive failure message to the `ExecutorClient` when actions fail, but be adding some more runtime failure info to a few other places as well.

PR #178, opened 2022-02-07
Change runtime failures to BT::NodeStatus::FAILURE, add logging info to ExecutorClient

sarcasticnature:
Also removed duplicate node pointer in `execute_action_node.cpp`
Solves: #174, #175
Contributes to: #161


plansys2_bt_actions/include/plansys2_bt_actions/BTActionNode.hpp
```diff
@@ -36,14 +36,11 @@ class BtActionNode : public BT::ActionNodeBase
     const std::string & xml_tag_name,
     const std::string & action_name,
     const BT::NodeConfiguration & conf)
-  : BT::ActionNodeBase(xml_tag_name, conf), action_name_(action_name)
+  : BT::ActionNodeBase(xml_tag_name, conf), action_name_(action_name),
+    return_failure_(false)
```
fmrico, line 40: Is it necessary to be a class member? `on_new_goal_received` could better return bool.

plansys2_bt_actions/include/plansys2_bt_actions/BTActionNode.hpp
```diff
@@ -134,32 +142,74 @@ class BtActionNode : public BT::ActionNodeBase
...
+        RCLCPP_ERROR(node_->get_logger(), "Could not create action client");
+        return BT::NodeStatus::FAILURE;
+      }
 
       // setting the status to RUNNING to notify the BT Loggers (if any)
       setStatus(BT::NodeStatus::RUNNING);
 
       // user defined callback
       on_tick();
+      if (return_failure_) {
```
fmrico, line 173: Why to check here? `on_tick` could change it? In this case, it could also return bool

plansys2_bt_actions/include/plansys2_bt_actions/BTActionNode.hpp
```diff
@@ -134,32 +142,74 @@ class BtActionNode : public BT::ActionNodeBase
...
+        return BT::NodeStatus::FAILURE;
+      }
 
       on_new_goal_received();
+      if (return_failure_) {
```
fmrico, line 178: the same here

plansys2_bt_actions/include/plansys2_bt_actions/BTActionNode.hpp
```diff
@@ -134,32 +142,74 @@ class BtActionNode : public BT::ActionNodeBase
...
+        cancel_goal();
+        return BT::NodeStatus::FAILURE;
+      }
     }
 
     // The following code corresponds to the "RUNNING" loop
     if (rclcpp::ok() && !goal_result_available_) {
       // user defined callback. May modify the value of "goal_updated_"
       on_wait_for_result();
+      if (return_failure_) {
```
fmrico, line 188: the same here?

plansys2_bt_actions/include/plansys2_bt_actions/BTActionNode.hpp
```diff
@@ -134,32 +142,74 @@ class BtActionNode : public BT::ActionNodeBase
...
+        if (return_failure_) {
+          cancel_goal();
+          return BT::NodeStatus::FAILURE;
+        }
       }
 
       rclcpp::spin_some(node_->get_node_base_interface());
 
+      // check if a derived class has set return_failure_ in a callback
+      if (return_failure_) {
```
fmrico, line 208: Why also check here?

plansys2_executor/src/plansys2_executor/ExecutorClient.cpp
```diff
@@ -80,10 +81,53 @@ ExecutorClient::execute_and_check_plan()
...
+          node_->get_logger(), "Plan failed due to a nullptr in the result");
+      } else if (result_.result->success) {
         RCLCPP_INFO(node_->get_logger(), "Plan Succeeded");
       } else {
-        RCLCPP_INFO(node_->get_logger(), "Plan Failed");
+        RCLCPP_ERROR(node_->get_logger(), "Plan Failed");
+        for (auto msg : result_.result->action_execution_status) {
+          switch (msg.status) {
+            case plansys2_msgs::msg::ActionExecutionInfo::SUCCEEDED:
+              RCLCPP_WARN_STREAM(
```
fmrico, line 94: Why WARN and not INFO or DEBUG?

plansys2_bt_actions/include/plansys2_bt_actions/BTActionNode.hpp
```diff
@@ -134,32 +142,74 @@ class BtActionNode : public BT::ActionNodeBase
...
+      // Get the required items from the blackboard
+      unsigned int server_timeout_int;
+      if (!getInput<unsigned int>("server_timeout", server_timeout_int)) {
+        // This will only happen if `providedPorts` is overridden and
+        // the child class does not provide the "server_timeout" port.
+        // Child classes can use the `providedBasicPorts` method to avoid
+        // this issue
+        RCLCPP_INFO(
+          node_->get_logger(),
+          "Missing input port [server_timeout], "
```
fmrico, line 154: I would rather prefer to join the messages to drain the 100 characters, instead of occupying more lines. The same in the other RCLCPP_* of this PR

fmrico:
Hi @sarcasticnature 

Could you add some tests to cover and verify all your contributions?

Thanks!!

plansys2_bt_actions/include/plansys2_bt_actions/BTActionNode.hpp
```diff
@@ -82,7 +78,11 @@ class BtActionNode : public BT::ActionNodeBase
   {
     BT::PortsList basic = {
       BT::InputPort<std::string>("server_name", "Action server name"),
-      BT::InputPort<std::chrono::milliseconds>("server_timeout")
+      BT::InputPort<unsigned int>(
```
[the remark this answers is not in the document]
sarcasticnature: There isn't an automatic conversion to `std::chrono::milliseconds` from a blackboard value (by my understanding), so a template specialization for the BehaviorTree.CPP function `convertFromString` would have to be implemented to specify a value in the xml, see [here](https://www.behaviortree.dev/tutorial_03_generic_ports/) for more details.

I see no reason to do this since we can just switch the type, but that is a point of opinion.

plansys2_bt_actions/include/plansys2_bt_actions/BTActionNode.hpp
sarcasticnature: if we check after spinning, a derived class can set `return_failure_` in a ROS callback, e.g. from a timer, subscriber, or the `on_feedback` function

plansys2_executor/src/plansys2_executor/ExecutorClient.cpp
sarcasticnature: that is an error, I will change it.  Thoughts on INFO vs DEBUG?

sarcasticnature:
The rationale for utilizing a member variable (`return_failure`) to signal a failure is that it will not break existing implementations, i.e. people who are already deriving from the class are not required to change their code despite the new feature.

plansys2_bt_actions/include/plansys2_bt_actions/BTActionNode.hpp
sarcasticnature: What do you mean by "join the messages"?

sarcasticnature:
I am willing and able to add tests, but am not well practiced at writing them.  I'll request that you take a close look at them once completed to make sure that they are written properly

fmrico:
> I'll request that you take a close look at them once completed to make sure that they are written properly

Sure!!

sarcasticnature:
Actually, if I run `colcon test` with the arguments specified in the GitHub action, I get a ton of test failures in multiple packages (none of which I touched).  The first test case of `plansys2_bt_actions` (`load_plugins`) crashes before it even gets to the new test cases.

If, however, I run the tests on `bt_action_test` only (`colcon test --event-handlers console_cohesion+ --return-code-on-test-failure --packages-select plansys2_bt_actions`) everything passes.

@fmrico thoughts?

sarcasticnature:
the errors with the other packages seem to be caused by running tests in parallel, everything seems to work with `--parallel-workers 1` or `--executor sequential`

still no clue why the `on_feedback` test is failing though

fmrico:
Yes, @sarcasticnature , you need to add `--parallel-workers 1`. Tests on some packages can interfere with tests in other ones.

sarcasticnature:
it looks like the tests passed this time without any changes to them... another point toward there being a race condition in the ROS action feedback/result.

@fmrico could you give the tests a shot on your machine?

fmrico:
I have tested it locally, and everything seems to be working well!!

Thanks @sarcasticnature

Issue #185, opened 2022-02-15
master do not compile on foxy


roveri-marco:
I've done a git pull on my previously downloaded version on the master, and the compilation failed. The reason being a missing include and the use of a method for the logger that is only supported in galactic or latest versions of ROS2 of rclcpp. In the patch, there are the changes I had to apply to have it compile successfully.

[compile.txt](https://github.com/IntelligentRoboticsLabs/ros2_planning_system/files/8072764/compile.txt)

Would it be possible to have a conditional compilation for the set_level method in the logger?

I belive that If galactic is needed, then it shall be properly stated.

I'm using foxy updated from the official repo in binary form, and running under Ubuntu 20.04.3 LTS, with g++ 9.3.0 (Ubuntu 9.3.0-17ubuntu1~20.04)

sarcasticnature:
This issue is relevant: #179

roveri-marco:
Although the <iomanip> seems to be not needed in galactic, from a programming point of view, the inclusion of needed header for used methods is a good practice instead of relying on other packages including them, especially if they are standard libraries.

As far as the set_level, I understand that now the master is for galactic, however, since it is a simple call to enable debugging it seems a minor change, and it might be conditioned to be in the galactic branch of ROS2 and disable in the other case, or using the proper code to enable that. 

These are just suggestions, and of course you are free to consider them or ignore thus closing this issue.

fmrico:
Hi all,

As said in #179, `master` points to `Galactic`. But, I understand that removing only this call, which is not critical, could make it compile also in Foxy. I will be happy to accept a PR only removing this call. Please, @roveri-marco , instead of attaching patch files, contribute following the usual way, as I described some time ago in https://github.com/IntelligentRoboticsLabs/ros2_planning_system/issues/168#issuecomment-994393296 .

Best
Francisco

roveri-marco:
Hi Francisco and all,

Sure I'll do as suggested in the forthcoming days.

Best
Marco

fmrico:
Fixed in #187

PR #187, opened 2022-02-17
Fixed compilation of main (working on galactic) to compile also on foxy

roveri-marco:
This PR will fix issue #185 by adding a missing include and commenting out a debugging code that prevented the compilation on foxy. All the tests with `colcon test` are passing.

### DIFF
--- a/plansys2_tools/src/plansys2_logger/LoggerNode.cpp
+++ b/plansys2_tools/src/plansys2_logger/LoggerNode.cpp
@@ -54,7 +54,6 @@ LoggerNode::LoggerNode()
   executing_plan_ = create_subscription<plansys2_msgs::msg::Plan>(
     "executing_plan", rclcpp::QoS(100).transient_local(),
     std::bind(&LoggerNode::executing_plan_callback, this, _1));
-
 }
 
 void
--- a/plansys2_tools/src/plansys2_logger/LoggerNode.cpp
+++ b/plansys2_tools/src/plansys2_logger/LoggerNode.cpp
@@ -15,6 +15,7 @@
 
 #include <string>
 #include <memory>
+#include <iomanip>
 
 #include "plansys2_msgs/msg/knowledge.hpp"
 #include "plansys2_msgs/msg/action_execution_info.hpp"
@@ -54,7 +55,7 @@ LoggerNode::LoggerNode()
     "executing_plan", rclcpp::QoS(100).transient_local(),
     std::bind(&LoggerNode::executing_plan_callback, this, _1));
 
-  get_logger().set_level(rclcpp::Logger::Level::Debug);
+  // get_logger().set_level(rclcpp::Logger::Level::Debug);
 }
 
 void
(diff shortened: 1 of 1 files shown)


plansys2_tools/src/plansys2_logger/LoggerNode.cpp
```diff
@@ -15,6 +15,7 @@
 
 #include <string>
 #include <memory>
+#include <iomanip>
```
fmrico, line 18: This include is not necessary

plansys2_tools/src/plansys2_logger/LoggerNode.cpp
```diff
@@ -54,7 +55,7 @@ LoggerNode::LoggerNode()
     "executing_plan", rclcpp::QoS(100).transient_local(),
     std::bind(&LoggerNode::executing_plan_callback, this, _1));
 
-  get_logger().set_level(rclcpp::Logger::Level::Debug);
+  // get_logger().set_level(rclcpp::Logger::Level::Debug);
```
fmrico, line 58: Remove instead of commenting it

fmrico (changes requested):

plansys2_tools/src/plansys2_logger/LoggerNode.cpp
sarcasticnature: I believe it is on foxy for some reason (default C++ version?), but @roveri-marco can confirm/deny

roveri-marco: I needed otherwise it complains for setprecision being undefined. 
It is not a matter of C++ version, it might be that in galactic it has been added somewhere else and thus it is not needed, but from a clean C++ perspective if you need to use the setprecision you need that include. Thus, I would push to have it.

plansys2_tools/src/plansys2_logger/LoggerNode.cpp
roveri-marco: Ok I'll do!

plansys2_tools/src/plansys2_logger/LoggerNode.cpp
fmrico: ok

fmrico (approved):

roveri-marco:
@fmrico I'm not sure the failure is due to the changes I did! Indeed it fails parsing a domain file that I've not changed! It seems related to other issue I've reported about numeric fluents/operations.

fmrico:
Run tests locally. The error seems to be:

```
.xml
  3: /home/runner/work/ros2_planning_system/ros2_planning_system/ros_ws/src/571sssz6sbp/ros2_planning_system/plansys2_tools/src/plansys2_logger/LoggerNode.cpp:57:  Redundant blank line at the end of a code block should be deleted.  [whitespace/blank_line] [3]
```

roveri-marco:
I've fixed removing the empty line. Funny thing that tests were not failing on my  computer.
What's the command you run to get this error?

I believe it would be good to specify precisely the commands to run before committing and asking for PR to avoid these iterations.

fmrico:
Hi @roveri-marco 

I think that this error could be detected by running `colcon test`. A few times CI test output is different from the local tests. Rare,  but it happens sometimes.

> What's the command you run to get this error?

I have gone to the failing workflow on this page and manually parse, at the end of each package, the failing tests. Sometimes errors are part of the expected behavior of one test, so go directly to the summary at the end of the packages.


CI SUCCESS!! Merging!! Thanks @roveri-marco

Issue #184, opened 2022-02-09
Possible problem in execution and in BT builder


roveri-marco:
We have a small example with a PDDL domain where there are some durative actions (e.g. `pick_up`, `move_carrier`, `deliver`). See the
`domain.pddl` file in attach. We created the CPP code for the above actions following examples in the distribution (we modified the code for `plansys2_simple_example`, and you can find attached also the resulting small code.

We compiled the code as suggested in the tutorial, and the we execute it with the following command.

`ros2 launch assignment assignment_launch.py`

On another terminal, we activate the plansys2 terminal with the command:

`ros2 run plansys2_terminal plansys2_terminal`

Then we specify the following problem

```
set instance robot1 robot
set instance carrier1 carrier
set instance p1 person
set instance p2 person
set instance c1 crate
set instance c2 crate
set instance depot location
set instance l1 location
set instance medicine content
set instance food content

set predicate (at_robot robot1 depot)
set predicate (at_carrier carrier1 depot)
set predicate (at_person p1 l1)
set predicate (at_person p2 l1)
set predicate (at_crate c1 depot)
set predicate (at_crate c2 depot)
set predicate (contain c1 medicine)
set predicate (contain c2 food)
set predicate (available c1)
set predicate (available c2)
set predicate (need p1 medicine)
set predicate (need p2 food)
set predicate (free robot1)
set function (= (capacity carrier1) 4)

set goal (and (have p1 medicine) (have p2 food))
```

If we ask for the generation of a plan with `popf` I get the following plan

```
0       (pick_up robot1 carrier1 c1 depot)      2
2.001   (pick_up robot1 carrier1 c2 depot)      2
4.001   (move_carrier robot1 carrier1 depot l1) 4
8.001   (deliver robot1 carrier1 c2 p2 l1 food) 1
9.002   (deliver robot1 carrier1 c1 p1 l1 medicine) 1
```

If I run it the execution succeeds, but on the first terminal I notice several errors in executing the `(pick_up robot1 carrier1 c2
depot)` and the `(deliver robot1 carrier1 c1 p1 l1 medicine)`. I believe there is a problem in the BT that it has been generated (also reported in attach). Indeed, from what I understand 

`(pick_up robot1 carrier1 c1 depot)` and `(pick_up robot1 carrier1 c2 depot)` are the first actions that start since they do not wait for anything

[...]

PR #190, opened 2022-03-03
action-graph-fix: Using std::list to create action graph rather than …

jjzapf:
There appears to be a small bug in the plan to action graph code. I first noticed this bug when running a “road trip” type scenario where the vehicle is supposed to visit various waypoints. The system would incorrectly create parallel actions. Specifically, a move and a visit action would be scheduled to occur simultaneously. This should not be possible because the move action has an “(at start (not (car_at ?s ?wp1)))” effect, while the visit action has an “(over all (car_at ?s ?wp1))” condition.

Examining the code, I discovered that the planning graph is currently created using a std::set type. However, a std::set does not maintain insertion order. So, graph traversal is not guaranteed to occur in the same order as the graph creation. This appears to lead to some weird consequences when performing the pruning operations. I switched to using a std::list type and I am no longer getting the incorrect parallel actions.

I will create a separate PR to add an action graph generation test using the “road trip” scenario I described above. Without the changes in this PR, that test should fail (assuming your computer behaves the same way as mine). With the changes in this PR, it should pass.

### DIFF
--- a/plansys2_executor/include/plansys2_executor/BTBuilder.hpp
+++ b/plansys2_executor/include/plansys2_executor/BTBuilder.hpp
@@ -57,8 +57,8 @@ struct GraphNode
   std::vector<plansys2::Predicate> predicates;
   std::vector<plansys2::Function> functions;
 
-  std::set<GraphNode::Ptr> in_arcs;
-  std::set<GraphNode::Ptr> out_arcs;
+  std::list<GraphNode::Ptr> in_arcs;
+  std::list<GraphNode::Ptr> out_arcs;
 };
 
 struct Graph
--- a/plansys2_executor/src/plansys2_executor/BTBuilder.cpp
+++ b/plansys2_executor/src/plansys2_executor/BTBuilder.cpp
@@ -318,7 +318,7 @@ BTBuilder::prune_backwards(GraphNode::Ptr new_node, GraphNode::Ptr node_satisfy)
   auto it = node_satisfy->out_arcs.begin();
   while (it != node_satisfy->out_arcs.end()) {
     if (*it == new_node) {
-      (*it)->in_arcs.erase(*it);
+      (*it)->in_arcs.remove(*it);
       it = node_satisfy->out_arcs.erase(it);
     } else {
       ++it;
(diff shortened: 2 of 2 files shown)


jjzapf:
Hi @sarcasticnature.

I'm not sure if this will solve #184 or not. However, the symptoms described in that issue do seem similar to the issues I described above. Unfortunately, the action_executor_client test in executor_test.cpp of the plansys2_executor package appears to be failing with these changes. The strange thing is that when I test locally on my machine all the tests pass. Note that to test this locally, I'm using the command

colcon test --packages-select plansys2_executor; colcon test-result --verbose --test-result-base build/plansys2_executor

This is a little bit difficult to debug, since I can't replicate the failure on my machine. Hopefully I will have some time to dig into this a bit more next week. If you want to test this mod out on your end I'd appreciate any help/suggestions. Thanks.

sarcasticnature:
I'll try to give it a shot this weekend.  I'm also having issues with the tests passing locally and failing in CI (with my recent PR), so you're not alone there 🙃

jjzapf:
I think this may just be a timeout issue. I modified the timeout to be 300 on a test branch in my GitHub space and the test passed. I will update the branch here. ... Hopefully it will also pass.

jjzapf:
@fmrico and @sarcasticnature, it looks like it was the timeout issue I mentioned above. I think this is ready for review now.

fmrico:
Hi @jjzapf 

I am not sure that this is the problem. I used `std::set` because I didn't consider the order relevant and because I could insert in the set several times the same node, without duplications. 

If you continue thinking that the order is relevant and you see improvements in using lists, I have no problem in considering this PR, but you should check before every push back that the data to insert is not already in the list.

Best

jjzapf:
Hi @fmrico

I also expected the order not to matter and was therefore surprised that this small change seemed to fix my problem.

I've added checks to look for existing links to avoid adding duplicates as you suggested. I don't think that these changes will add any regressions and they seem to help at least empirically. However, you may indeed be right that this does not fix the real problem. I'm fine with whatever you decide regarding whether or not to merge these changes.

If I find some time, I'll try to dig into the plan -> action graph -> behavior tree algorithm to see if there's anything else I can find.

PR #191, opened 2022-03-03
action-graph-test: Adding unit test to verify action graph generation.

jjzapf:
This PR adds an action graph generation test to the plansys2_executor package. This PR should be merged after #190. Without the changes in that PR, the new test should fail. With the changes in that PR, the new test should pass. See #190 for a detailed description of the current bug that this test was designed to tease out.


jjzapf:
It appears that the PR that should have passed the checks did not and the PR that should not have passed the checks did. ... In the words of Homer Simpson, D'oh! ... I haven't seen this locally, but I will do some more testing.

fmrico:
Thanks @jjzapf 

Let me know when it is ready :wink:

jjzapf:
Hi @fmrico.

I think this is ready to go. My guess is that the issue I described in #190 (i.e. std::set does not maintain insertion order) leads to some amount of randomness when running on different machines. This may be why the new test is passing here but not on my local machine. The hope is that #190 should fix the issue for all machines as it should remove the randomness.

When running the new test on my machine (without the changes in #190), I get the following error.

    0, 0, 0, move herbie wp0 wp1
    0, 2, 2, move herbie wp1 rp0
    0, 3, 3, refuel herbie rp0
    0, 4, 4, move herbie rp0 wp3
    0, 5, 5, visit herbie wp3
    0, 6, 6, move herbie wp3 wp2
    0, 7, 7, visit herbie wp2
    0, 1, 1, visit herbie wp1
    /home/jjzapf/src/NIWCMM-371/src/niwc/ros2_planning_system/plansys2_executor/test/unit/btbuilder_tests.cpp:908: Failure
    Expected equality of these values:
      std::get<1>(tabulated_graph[i])
        Which is: 2
      std::get<1>(expected_graph[i])
        Which is: 1
    terminate called without an active exception

The printed table is formed by traversing the action graph. Note that the (visit herbie wp1) action is printed last when it should be printed second. This means that it was the last node visited in the depth first search rather than the second. This implies that there was a split after the first node. That is, the (move herbie wp0 wp1) node branches to (move herbie wp1 rp0) and (visit herbie wp1). However, these actions cannot be performed in parallel. Applying the changes in #190 resolves this issue on my machine.

You might try running this test on your machine with and without the changes in #190 to see if your machine behaves similar to mine.

PR #193, opened 2022-03-03
Increase default server tiemout

fmrico:
Hi,

Last merges, which tests passed (both in local and in CI separately), makes the tests fail in master. This PR is to investigate and to fix this.

Best


sarcasticnature:
I believe it's due to the race condition I have been speculating about in my other PR.  I think what is happening is that the ROS action server is sending a goal result without ever sending feedback, which is causing the test to fail

sarcasticnature:
add a delay in the test server to allow time for the feedback to send before it returns success.  I can't work on a fix at this moment, but should be able to knock something out tonight

fmrico:
Tests in PR works, but not in master... :thinking:

sarcasticnature:
The line you changed increased the amount of time the node will wait for an action server to be available, which does not add any delay.  I may have a fix but will need to create a PR to see if it passes CI

PR #196, opened 2022-03-05
Fix run entire plan

fmrico:
Hi,

#192 introduced a small bug that does not let to run the entire map only by typing `run`. This PR fixes it.

Best


fmrico:
Failing for reasons not related to this PR. Mixing as this is a simple fix.

PR #197, opened 2022-03-05
Refactor BTActionNode

fmrico:
Hi,

This PR tries to solve issues related to the merge of #178, not solved with #193, and related to #195 .
This is a refactor to #178 and a revert of some functionality that I think is necessary.

An important change in this PR is the `on_tick()` method. Now it is called (as it should be from initial) at each tick, and return a `BT::NodeStatus` that overrides `tick()` operation.

Take a look at this, @sarcasticnature , please
Best

### DIFF
--- a/plansys2_bt_actions/include/plansys2_bt_actions/BTActionNode.hpp
+++ b/plansys2_bt_actions/include/plansys2_bt_actions/BTActionNode.hpp
@@ -70,7 +70,18 @@ public:
 
     // Make sure the server is actually there before continuing
     RCLCPP_INFO(node_->get_logger(), "Waiting for \"%s\" action server", action_name.c_str());
-    return action_client_->wait_for_action_server(server_timeout_);
+
+    bool success_waiting = action_client_->wait_for_action_server(server_timeout_);
+
+    if (!success_waiting) {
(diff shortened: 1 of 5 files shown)


sarcasticnature:
@fmrico looks good to me. Changing `on_tick` to return a `BT::NodeStatus` was my original plan, but I did not want to break backwards compatibility; that decision is yours to make as the maintainer.

I see you aren't messing around anymore with the `on_feedback` race condition in the test, lol (I like the fix :shipit:).

**There is something important for you do do after this change, however:** the [documentation](https://intelligentroboticslab.gsyc.urjc.es/ros2_planning_system.github.io/tutorials/docs/bt_actions.html) needs to be updated to match!  Same goes for the [examples repo](https://github.com/IntelligentRoboticsLabs/ros2_planning_system_examples).  I would also suggest that you make an announcement on ROS Discourse the next time you sync the galactic branch, as this will break all code that utilizes this class

I am willing to help with updating things if needed.

fmrico:
Hi @sarcasticnature 

Oh, I have just updated it. Thanks for your offering. I saw it late :disappointed: 

If you find something else to fix, warn me, or directly open a PR.

Thanks!!

PR #198, opened 2022-03-05
Check reqs

### LINKED ISSUE #184: Possible problem in execution and in BT builder

fmrico:
Hi

This PR is related to #191 and fixes #184.

When creating the graph, the roots (the actions to initiate an execution flow) are identified. They were not correctly identified, because not all the cases were checked to determine if these roots were compatible. Now, if an action requires a predicate which is an effect of another root, or an action produces an effect which is a requisite of another root, it is not considered a root.

Best
Francisco

### DIFF
--- a/plansys2_executor/src/plansys2_executor/BTBuilder.cpp
+++ b/plansys2_executor/src/plansys2_executor/BTBuilder.cpp
@@ -373,7 +373,6 @@ BTBuilder::get_graph(const plansys2_msgs::msg::Plan & current_plan)
   auto functions = problem_client_->getFunctions();
 
   graph->roots = get_roots(action_sequence, predicates, functions, node_counter);
-  std::cerr << "Roots = " << graph->roots.size() << std::endl;
 
   // Apply root actions
   for (auto & action_node : graph->roots) {
--- a/plansys2_executor/src/plansys2_executor/BTBuilder.cpp
+++ b/plansys2_executor/src/plansys2_executor/BTBuilder.cpp
@@ -220,20 +220,39 @@ BTBuilder::is_parallelizable(
   const plansys2::ActionStamped & action,
   const std::list<GraphNode::Ptr> & ret) const
 {
-  std::vector<plansys2_msgs::msg::Node> action_at_start_requirements;
-  parser::pddl::getPredicates(action_at_start_requirements, action.action->at_start_requirements);
+  std::vector<plansys2_msgs::msg::Node> action_requirements;
+  std::vector<plansys2_msgs::msg::Node> action_effects;
+  parser::pddl::getPredicates(action_requirements, action.action->at_start_requirements);
+  parser::pddl::getPredicates(action_requirements, action.action->over_all_requirements);
+  parser::pddl::getPredicates(action_requirements, action.action->at_end_requirements);
+  parser::pddl::getPredicates(action_effects, action.action->at_start_effects);
+  parser::pddl::getPredicates(action_effects, action.action->at_end_effects);
 
   for (const auto & other : ret) {
-    std::vector<plansys2_msgs::msg::Node> other_over_all_requirements;
-    parser::pddl::getPredicates(
-      other_over_all_requirements,
-      other->action.action->over_all_requirements);
+    std::vector<plansys2_msgs::msg::Node> other_requirements;
+    std::vector<plansys2_msgs::msg::Node> other_effects;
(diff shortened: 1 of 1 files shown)


jjzapf:
Hi @fmrico.

The changes you made to the is_parallelizable function make sense to me. Basically you are testing if the effects of a new action contradict the requirements of an existing action in the input set. Or likewise, you are testing if the effects of an existing action contradict the requirements of a new action.

This function is only used to check if actions can be run in parallel at the root level, correct? Shouldn't we also be checking if actions can be run in parallel throughout the rest of the graph creation? In the example I posted in PR #191, the incorrect parallel branches happen after the root node.

One other note ... The is_parallelizable function currently only checks the predicates. In the future we should also check for function effect/requirement contradictions.

Thanks for addressing this issue! I will try to offer whatever support I can.

Issue #194, opened 2022-03-03
Features request


roveri-marco:
This is not a bug, but a feature request. 

It would be worth having the possibility to upload a plan generated externally to the plansy2 infrastructure for then executing it.
From the plansys2 terminal, this will simplify interaction.  This for instance is very useful to validate certain scenarios without having to specify the instances/predicates/functions/goals, call the planner and then execute the plan. In this case, one can simply specify instances/predicates/functions and then can load a plan, and execute it without calling the planner.
But also from the overall architecture, I think that having the possibility to send a plan manually specified to the system is an interesting use case.

Another interesting feature for the terminal would be to load a set of commands from a file. With this feature, it is much easier to deal with the terminal. Indeed, one can write the commands in the file, and simply load the script without having to copy-paste the commands from the file to the terminal.

I will be willing to work directly on these features (especially the second)

fmrico:
Hi @roveri-marco 

> It would be worth having the possibility to upload a plan generated externally to the plansy2 infrastructure for then executing it.
> From the plansys2 terminal, this will simplify interaction. This for instance is very useful to validate certain scenarios without having to specify the instances/predicates/functions/goals, call the planner and then execute the plan. In this case, one can simply specify instances/predicates/functions and then can load a plan, and execute it without calling the planner.
> But also from the overall architecture, I think that having the possibility to send a plan manually specified to the system is an interesting use case.

I think that most of this functionality is already there. I did something similar in #192. Maybe we could extend [ExecutePlan.action](https://github.com/IntelligentRoboticsLabs/ros2_planning_system/blob/16b2dee52faa3bca1dfaf1ac8824bf0a8ac496a4/plansys2_msgs/action/ExecutePlan.action) with an optional entry in the request with the plan. If it is void, ask a plan to the planner.

> Another interesting feature for the terminal would be to load a set of commands from a file. With this feature, it is much easier to deal with the terminal. Indeed, one can write the commands in the file, and simply load the script without having to copy-paste the commands from the file to the terminal.

Well, you can do it by using the input redirection operator '<' in the shell

Best
Francisco

roveri-marco:
@fmrico I'll check #192 and I'll see...

For the second part It is not enough since if you would like then to continue interacting with the terminal it is not easy.
I've already finished implementing this last feature and I'm testing it. I've also added an help command that I believe it is useful for interacting with the terminal. I'll add a PR soon.

Best
Marco

[...]

PR #200, opened 2022-03-09
Extend terminal with supporting commands

roveri-marco:
This PR addresses #194 by adding 
- a help command
- the source command to load commands from a file
- extended run command with planfrom to read a file from a file
- added tests for the source/extended run (all the tests pass)

### DIFF
--- a/plansys2_terminal/src/plansys2_terminal/Terminal.cpp
+++ b/plansys2_terminal/src/plansys2_terminal/Terminal.cpp
@@ -78,15 +78,18 @@ char * completion_generator(const char * text, int state)
   static std::vector<std::string> matches;
   static size_t match_index = 0;
 
-  std::vector<std::string> vocabulary{"get", "set", "remove", "run", "check"};
+  std::vector<std::string> vocabulary{"get", "set", "remove", "run",
+    "check", "source", "help", "?", "quit"};
   std::vector<std::string> vocabulary_check{"actors"};
   std::vector<std::string> vocabulary_set{"instance", "predicate", "function", "goal"};
   std::vector<std::string> vocabulary_get{"model", "problem", "domain", "plan"};
   std::vector<std::string> vocabulary_remove{"instance", "predicate", "function", "goal"};
-  std::vector<std::string> vocabulary_run{"action", "num_actions", "planfrom"};
+  std::vector<std::string> vocabulary_run{"action", "num_actions", "plan-file"};
   std::vector<std::string> vocabulary_get_problem{"instances", "predicates", "functions", "goal"};
   std::vector<std::string> vocabulary_get_model{"types", "predicates", "functions", "actions",
     "predicate", "function", "action"};
+  // The help is initialized with all possible commands in the vocabulary
+  std::vector<std::string> vocabulary_help(vocabulary);
 
   if (state == 0) {
     // During initialization, compute the actual matches for 'text' and keep
@@ -114,6 +117,8 @@ char * completion_generator(const char * text, int state)
           current_vocabulary = &vocabulary_run;
         } else if (current_text[0] == "check") {
           current_vocabulary = &vocabulary_check;
+        } else if ((current_text[0] == "help") || (current_text[0] == "?")) {
+          current_vocabulary = &vocabulary_help;
         }
       } else if (current_text.size() == 3) {
         if (current_text[0] == "get" && current_text[1] == "problem") {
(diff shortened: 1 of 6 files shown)


plansys2_terminal/src/plansys2_terminal/Terminal.cpp
```diff
@@ -82,7 +83,7 @@ char * completion_generator(const char * text, int state)
   std::vector<std::string> vocabulary_set{"instance", "predicate", "function", "goal"};
   std::vector<std::string> vocabulary_get{"model", "problem", "domain", "plan"};
   std::vector<std::string> vocabulary_remove{"instance", "predicate", "function", "goal"};
-  std::vector<std::string> vocabulary_run{"action", "num_actions"};
+  std::vector<std::string> vocabulary_run{"action", "num_actions", "planfrom"};
```
fmrico, line 86: - Add also `help` to completion, with all the arguments
- `planfrom` is not bad (I prefect `plan-from`), but maybe `plan-file` is better.

plansys2_terminal/src/plansys2_terminal/Terminal.cpp
```diff
@@ -823,11 +868,29 @@ Terminal::process_run(std::vector<std::string> & command, std::ostringstream & o
...
         os << e.what() << " with arg: [" << command[0] << "]" << std::endl;
       }
+    } else if (command[0] == "planfrom") {
+      if (command.size() == 2) {
+        std::optional<plansys2_msgs::msg::Plan> plan = parse_plan(command[1]);
+        if (!plan.has_value()) {
+          os << "Plan could not be loaded " << std::endl;
+          return;
+        }
+        try {
```
fmrico, line 878: What exception can be thrown here?

plansys2_terminal/src/plansys2_terminal/Terminal.cpp
```diff
@@ -891,6 +954,113 @@ Terminal::process_check(std::vector<std::string> & command, std::ostringstream &
...
+  if (cmd != cmds.end()) {
+    os << cmd->first << std::endl;
+    os << cmd->second << std::endl;
+    return;
+  }
+  os << "Command \"" << command[0] << "\" command not found" << std::endl;
+}
+
+void
+Terminal::process_source(std::vector<std::string> & command, std::ostringstream & os)
```
fmrico, line 1023: I would prefer to avoid this command. I still think that the effect of this command is similar to opening another terminal and run  `ros2 run plansys2_terminal plansys2_terminal < file_with_commands`. This part only adds complexity to the terminal, with variables set inside and outside the function, prone to lateral effects.

plansys2_terminal/src/plansys2_terminal/Terminal.cpp
roveri-marco: Sure, I've forgot!

plansys2_terminal/src/plansys2_terminal/Terminal.cpp
roveri-marco: You may simulate this with yet another terminal, fine but if you're interacting over the terminal, and you would like to continue interacting with the terminal, instead of opening another terminal, set up the ros envirionemt, and run the command, you can do the very same thing from one single terminal.

To be honest, playing with the terminal I found this feature very useful! We can think to a better mechanism instead of variables, but I strongly believe that having this feature is useful and the "complexity" added is manageable.

plansys2_terminal/src/plansys2_terminal/Terminal.cpp
roveri-marco: Fixed in the latest push!

plansys2_terminal/src/plansys2_terminal/Terminal.cpp
fmrico: Ok, let's add his feature to the terminal, but we have to find a better implementation that does not use global variables, like `inside_source` and `finish_parsing`

roveri-marco: I'm trying an implementation without these global variables, and I'll commit soon as all the test passes

roveri-marco: I've made the code reentrant by removing global variables.

All tests pass.

fmrico:
Merged!! Thanks @roveri-marco 😃

antbono:
@roveri-marco What is the terminal command to read a problem.pddl file?
Thanks

PR #201, opened 2022-03-09
Generalize is parallelizable

jjzapf:
The is_parallelizable function in BTBuilder currently only checks for contradictions between predicate effects and requirements. This PR will generalize the is_parallelizable function to work with functions as well.

### DIFF
--- a/plansys2_executor/include/plansys2_executor/BTBuilder.hpp
+++ b/plansys2_executor/include/plansys2_executor/BTBuilder.hpp
@@ -122,6 +122,8 @@ protected:
     std::vector<plansys2::Function> & functions) const;
   bool is_parallelizable(
     const plansys2::ActionStamped & action,
+    const std::vector<plansys2::Predicate> & predicates,
+    const std::vector<plansys2::Function> & functions,
     const std::list<GraphNode::Ptr> & ret) const;
 
   std::string get_flow_tree(
--- a/plansys2_executor/src/plansys2_executor/BTBuilder.cpp
+++ b/plansys2_executor/src/plansys2_executor/BTBuilder.cpp
@@ -287,9 +287,8 @@ BTBuilder::get_roots(
   auto it = action_sequence.begin();
   while (it != action_sequence.end()) {
     const auto & action = *it;
-    if (is_action_executable(
-        action, predicates,
-        functions) && is_parallelizable(action, predicates, functions, ret))
+    if (is_action_executable(action, predicates, functions) &&
+      is_parallelizable(action, predicates, functions, ret))
     {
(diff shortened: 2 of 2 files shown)


jjzapf:
Hold up on this one. I'm not sure if this approach is 100% correct for the functions case.

Right now I check if the effects of an action would contradict the requirements of another in the set by applying the effects and then checking the requirements. In the case of a function, however, the effect may modify the value of the function without nullifying a requirement on the function.

For example, suppose I have a function that keeps track of the number of marbles I have. An effect could reduce that number of marbles without nullifying a requirement that I have some minimum number of marbles.

I think that I also need to ensure that the resources (i.e. PDDL functions) are independent among the actions in the parallel set so that only one action operates on a given resource at a time. 

Does this make sense, or am I losing my marbles? ... Sorry, couldn't resist the joke. :)

jjzapf:
**Sharing Resources**
The more I think about this, I'm inclined to believe that in general parallel actions should be allowed to share resources. 
For example, 2 robots should be allowed to add/remove marbles from a bin at the same time. If the resource cannot be shared, the designer should encode the domain to ensure that the actions are mutually exclusive.

Would you agree with this @fmrico? Any concerns about concurrent writes?

**Checking for Contradictions at Any Point in Time**
I came across a special case of the elevator example in "An Introduction to the Planning Domain Definition Language". The example demonstrated the use of a "unary resource" in the following open-door action.

    (:durative-action open-door
     :parameters (?lift - elevator ?flr - num)
     :duration (= ?duration (door_speed ?lift))
     :condition (and (at start (not (open ?lift)))
                              (at start (can-stop-at ?flr))
                              (overall (lift-at ?lift ?flr)))
     :effect (and (at start (open ?lift))
                        (at start (not (can-stop-at ?flr)))
                        (at end (not (open ?lift)))
                        (at end (can-stop-at ?flr)))
    )

The predicate "can-stop-at restricts" the domain such that only one elevator can open its door at a particular floor at any time. Note the "at start" and "at end" effects involving this predicate. The second negates the first.

When computing a parallel set, I think this implies that we must check for contradictions at any possible point in time, since we don't know exactly how the actions may overlap. To do this, I think we need to first apply the "at start" effects and check all the requirements (i.e. "at start", "overall", and "at end"). If this passes, we should then apply the "at end" effects and again check all the requirements.

See recent changes. I think I am now checking for contradictions at all possible points of overlap.

fmrico:
Hi @jjzapf 

This part is critical to the correct behavior of PlanSys2. Let's stay alert if issues arise from this in the next weeks, but I think it is correct. I think you did a really good job here.

Merging!!

jjzapf:
Thanks! :)

PR #203, opened 2022-03-17
fix-getSubtrees: Fixing bug in getSubtrees function in Utils.cpp.

jjzapf:
I found a bug in the getSubtrees function in plansys2_pddl_parser/Utils.cpp. This function retrieves a set of PDDL subtrees from a parent tree. When creating a subtree object, the nodes are reindexed. Obviously, this requires that the children vector also be updated. In the current implementation, we push the new values onto the children vector without clearing the vector first. This results in a children vector with erroneous indices at the front. To correct this, we simply need to clear the children vector prior to pushing the new indices.

As far as I can tell this function is currently only used to retrieve subgoals in the Problem Expert. Most often these are simple predicates without any children. This may explain why we did not find this bug sooner.


fmrico:
Good catch!!

Merged

Issue #188, opened 2022-02-19
Numeric fluents predicates in :conditions  cause  exception 


roveri-marco:
If  the PDDL planning domain contains numernic fluents, and the domain contains an action with a numeric predicate  of the form `(=  <expr> <expr>)`  or other operators (e.g. >= <= ...) the PDDL parser throws and exception that then prevents to perform plan generation although the popf or tfd planner are generating a plan for that.

In attach there are two simple domain/problems that show this:

[check_domain.pddl.txt](https://github.com/IntelligentRoboticsLabs/ros2_planning_system/files/8118584/check_domain.pddl.txt)
[check_problem.pddl.txt](https://github.com/IntelligentRoboticsLabs/ros2_planning_system/files/8118585/check_problem.pddl.txt)

Indded, if one tries the following command:
`lib/plansys2_pddl_parser/parser check_domain.pddl.txt check_problem.pddl.txt`
you get the following output:
```
terminate called after throwing an instance of 'parser::pddl::UnknownToken'
  what():   does not name a known token
Aborted
```

Since it is now possible to use numeric fluents, one is also expecting that they are supported in preconditions.

Since POPF correctly deals with this file it is a pity it is supported.

```
popf2-clp check_domain.pddl.txt check_problem.pddl.txt 
Number of literals: 1
Constructing lookup tables:
Post filtering unreachable actions: 
None of the ground temporal actions in this problem have been recognised as compression-safe
Initial heuristic = 2.000
b (1.000 | 10.000);;;; Solution Found
; States evaluated: 3
; Cost: 10.000
; Time 0.00
0.000: (load r0 loc0)  [10.000]
```

@fmrico I can help to fix this, but I need a bit of advice on how to extend the parser to handle in the preconditions such kind of numeric predicates (and whenever needed).

roveri-marco:
It seems that the only thing that are accepted with the `=` operator are references to parameters e.g. for the action in the example one can write `(= ?rob ?rob)`,.

Moreover, if one declares a constant 
```(:constant
   rrr - robotic_agent
)
```
And then uses it in `(= ?rob rrr)`, the parser complains, as well as if one uses the constant  `rrr` later on in a precondition e.g. `(> (carrying_r rrr) (* (max_capacity) 0))`. 

One possibility would be to handle the `=` as the other operators (e.g. `<`), or depending on the arguments being them numeric or object/parameter to properly construct the respective expression/constant value.

[...]

PR #204, opened 2022-03-28
Possible fix for issue #188

roveri-marco:
This is  a possible fix for issue #188 
- Now it correctly parse `(= ?r1 ?r2)`, `(= ?r rob)`, `(= (numeric_fluent ?r) (numeric_fluent2 ?r1 ?r2))` where `rob` is a constant and `?r` , `?r1`, and '?r2' are action parameters. 
-  Added supports for constants in the arguments of numeric fluents (e.g. `(numeric_fluent rob ?r)`)
-  Added some informative output for not supported features (e.g. `#t` as used in PDDL+)
-  Added one test case for the successful parsing


jjzapf:
Still reviewing these changes, but over all they look good. Nice job!

The use of "==" to compare 2 doubles still seems a bit dubious to me. I'm not really sure about the best way to handle this. However, I did come across this approach, which might work.

https://en.cppreference.com/w/cpp/types/numeric_limits/epsilon

In that example, the comparison epsilon value is defined in terms of the machine epsilon value, the magnitude of the values, and the desired precision in ULPs. Perhaps we could just hard code the number of ULPs? (The example uses a value of 2.) Even with hard coding the number of ULPs, the comparison epsilon value will vary with the magnitude of the input values. This seems like it could strike the correct balance between problem generality and ease of use.

plansys2_pddl_parser/src/plansys2_pddl_parser/Utils.cpp
```diff
@@ -14,7 +14,7 @@
 
 #include <regex>
 #include <iostream>
-#include <functional> 
+#include <functional>
```
jjzapf, line 17: Is this include needed?

plansys2_pddl_parser/src/plansys2_pddl_parser/Expression.cpp
```diff
@@ -63,34 +71,90 @@ Expression * createExpression( Stringreader & f, TokenStruct< std::string > & ts
 		f.next();
 		std::string s = f.getToken();
 		if ( s == "+" || s == "-" || s == "*" || s == "/" ) {
```
jjzapf, line 73: Should we also check for these: "=", ">=", ">", "<=", "<" ?

plansys2_pddl_parser/src/plansys2_pddl_parser/Expression.cpp
```diff
@@ -63,34 +71,90 @@ Expression * createExpression( Stringreader & f, TokenStruct< std::string > & ts
...
+   // It is a composite expression
 			CompositeExpression * ce = new CompositeExpression( s );
 			ce->parse( f, ts, d );
 			return ce;
-		}
-		else {
+		}	else {
+   // It is a function expression
```
jjzapf, line 79: Unfortunately, most of the plansys2_pddl_parser files seem to use tabs instead of spaces for indentation. It's a bit hard to follow all the if, else if, and else statements in these changes as I believe you are using spaces. Could you use tabs to follow the original indentation scheme?

In the future, I would support replacing all the tabs in these files with spaces.

roveri-marco:
> 

Indeed, this could be a good approach in the internals, but at planning level while specifying the domain one is expected to use =. We may handle this case in the proper way in a second PR if ok for you, otherwise we can extend this to handle as suggested.

plansys2_pddl_parser/src/plansys2_pddl_parser/Expression.cpp
roveri-marco: If ok, I can change all the tabs to spaces since easier ;-) I was not noticing since this package is not subject to lint as for instance it is the case for the terminal.

plansys2_pddl_parser/src/plansys2_pddl_parser/Expression.cpp
roveri-marco: I'm not sure since here we are building expressions (e.g. `(= (* (+ a b) d)))`) and it is not possible to mix equality within expressions. The relational operators are handled above if I'm not mistaken.

plansys2_problem_expert/src/plansys2_problem_expert/Utils.cpp
```diff
@@ -200,6 +200,14 @@ std::tuple<bool, bool, double> evaluate(
               return std::make_tuple(true, false, 0);
             }
             break;
+          case plansys2_msgs::msg::Node::COMP_EQ:
+            if (std::get<2>(left) == std::get<2>(right)) {
```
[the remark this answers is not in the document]
roveri-marco: Ok, we can handle it as suggested then.

plansys2_pddl_parser/src/plansys2_pddl_parser/Expression.cpp
jjzapf: I'd be fine converting all tabs to spaces. Although, that might create a very large diff in this package. Perhaps it would be better to do that in a separate PR. We should see what @fmrico has to say about this.

fmrico: Yes, probably a separate PR is better. And, of, of course, spaces better tan tabs

fmrico:
LGTM!! Merging

> We may handle this case in the proper way in a second PR if ok for you, otherwise we can extend this to handle as suggested.
:+1:

PR #208, opened 2022-04-04
Plan dotgraph color bug

xydesa:
More accurately getting the status of an action by including the action's start time in the index for coloring the dotgraph.

*Testing*
* Basing the test off of the simple example in the ros2_planning_system_examples repository.
* Create a workspace from this repos file: 
[plansys2-dotgraph-bug.repos.txt](https://github.com/IntelligentRoboticsLabs/ros2_planning_system/files/8412136/plansys2-dotgraph-bug.repos.txt)
* Terminal 1: `ros2 launch plansys2_simple_example plansys2_simple_example_launch.py domain:=domain_charging`
* Terminal 2: `ros2 run plansys2_terminal plansys2_terminal --ros-args -p problem_file:=./install/plansys2_simple_example/share/plansys2_simple_example/pddl/charging_problem.pddl`
* Terminal 3: `ros2 run rqt_dotgraph rqt_dotgraph`
* After you type run and hit enter in Terminal 4 and the plan starts executing, you should see the plan in the rqt_dotgraph window. It should look something like this:
plansys2-dotgraph-bug-fixed
Without this fix, it will look like this:
plansys2-dotgraph-bug

### DIFF
--- a/plansys2_executor/include/plansys2_executor/BTBuilder.hpp
+++ b/plansys2_executor/include/plansys2_executor/BTBuilder.hpp
@@ -135,7 +135,7 @@ protected:
     GraphNode::Ptr node, std::shared_ptr<std::map<std::string,
     ActionExecutionInfo>> action_map, int level = 0);
   ActionExecutor::Status get_action_status(
-    std::shared_ptr<plansys2_msgs::msg::DurativeAction> action,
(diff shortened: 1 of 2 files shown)


fmrico:
Thanks @xydesa!!

Issue #212, opened 2022-04-05
Terminal: set predicate with no argument causes terminal to exit


roveri-marco:
Let's consider the simple example in the attachment where I have a predicate `(nobusy)`, and once compiled run 
* `ros2 launch pddl_project_problem_4 pddl_project_problem_4_launch.py` on one terminal
* `ros2 run plansys2_terminal plansys2_terminal`  on the second terminal

Then on the terminal you can type the commands in the `commands.txt` file. Once  the `set predicate (nobusy)` is issued, the terminal exit to the shell without any further possibility to issue commands but restart the terminal.

Files to reproduce the problems:
* [commands.txt](https://github.com/IntelligentRoboticsLabs/ros2_planning_system/files/8418128/commands.txt)
* [pddl_project_problem_4-buggy-no-args.zip](https://github.com/IntelligentRoboticsLabs/ros2_planning_system/files/8418130/pddl_project_problem_4-buggy-no-args.zip)

fmrico:
I think we an close this

PR #213, opened 2022-04-05
Fix for issue #212 - predicate with no args

roveri-marco:
Fixed the handling of predicates with no arguments.

All tests and coding rules passe.


fmrico:
Thanks @roveri-marco

Issue #211, opened 2022-04-05
Problem in the constructed BT structure and execution


roveri-marco:
I've created a simple example with three durative actions `move` `deliver` and `load`. I've added a mutual exclusion predicate `(nobusy ?r)' 
The model does not contain any numeric fluent but only predicates. I can successfully run the launcher and the terminal

`ros2 launch pddl_project_problem_4 pddl_project_problem_4_launch.py` on one terminal

`ros2 run plansys2_terminal plansys2_terminal`  on the second terminal

Then on the terminal I issue the commands specified in the command files in attach, and then I ask to generate a plan and to run it.

```
> get plan
plan: 
0:	(load r h c1 depot n3 empty)	[2]
2.001:	(load r h c4 depot n2 n3)	[2]
4.001:	(move h r depot l1)	[3]
7.001:	(deliver r h p1 c1 food l1 n2 n3)	[2]
9.001:	(move h r l1 depot)	[3]
12.001:	(load r h c2 depot n2 n3)	[2]
14.001:	(move h r depot l2)	[3]
17.001:	(deliver r h p2 c2 food l2 n2 n3)	[2]
19.001:	(move h r l2 depot)	[3]
22.001:	(load r h c3 depot n2 n3)	[2]
24.001:	(move h r depot l3)	[3]
27.001:	(deliver r h p4 c3 food l3 n2 n3)	[2]
29.002:	(deliver r h p4 c4 medecine l3 n3 empty)	[2]
> run
[(deliver r h p1 c1 food l1 n2 n3) [(move h r l1 depot) [ERROR] [1649160044.052233393] [executor_client]: Plan Failed
[WARN] [1649160044.052428155] [executor_client]: Action: (deliver r h p1 c1 food l1 n2 n3):7001 was cancelled
[WARN] [1649160044.052531583] [executor_client]: Action: (deliver r h p2 c2 food l2 n2 n3):17001 was not executed
[WARN] [1649160044.052605985] [executor_client]: Action: (deliver r h p4 c3 food l3 n2 n3):27001 was not executed
[WARN] [1649160044.052680407] [executor_client]: Action: (deliver r h p4 c4 medecine l3 n3 empty):29002 was not executed
[WARN] [1649160044.052755220] [executor_client]: Action: (load r h c1 depot n3 empty):0 succeeded with message_status: Loaded c1 on h
[...]

PR #216, opened 2022-04-12
Plan-to-Action Graph Bug Fix

jjzapf:
This pull request is in response to Issue #211.

To test whether 2 actions can be run in parallel, we (1) test whether each action is executable given the current state and (2) check whether the effects of one action will contradict the requirements of the other. This is good, but there is a problem in the current approach.

In the current approach, we check whether the "at end" effects of one action will contradict any of the conditions of the other action. However, @roveri-marco provided an example of a valid plan where two actions could be run in parallel in which an "at end" effect of one action contradicted an "at start" condition of the other. Logically, this makes sense. To test whether 2 actions can run in parallel, we should only need to check for contradictions between the "at start" effects and the "at start" and "over all" conditions. This PR fixes this bug

When working on the above bug, another bug was found in the prune_backwards function of BTBuilder. In this case, the superfluous parent nodes were not being correctly removed from the input arc set. This PR also fixes that bug.

### DIFF
--- a/plansys2_executor/include/plansys2_executor/BTBuilder.hpp
+++ b/plansys2_executor/include/plansys2_executor/BTBuilder.hpp
@@ -136,7 +136,7 @@ protected:
     GraphNode::Ptr node,
     std::list<std::string> & used_nodes,
     int level = 0);
-  std::string get_flow_dotgraph(GraphNode::Ptr node, int level = 0);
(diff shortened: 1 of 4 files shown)


fmrico:
Thanks @jjzapf 

I will merge this now. @roveri-marco , may you check that plans are correctly carried out with your examples.

Thank you!!

Issue #226, opened 2022-05-22
bt_action_node will never 'finish' if an exception is thrown in the provided BT


sarcasticnature:
As the title says, a `bt_action_node` will never call the `finish` function of `ActionExecutorClient` if an exception is thrown in the provided BT.  That means that the `Executor` will hang indefinitely, as no failure message will ever be passed back through `actions_hub`.  Unfortunately, it is not uncommon for this to occur in the form of a `BT::RuntimeError` or `BT::LogicError` thrown by a BT node, e.g. if a port value can not be read or properly set.

I propose that a `try`/`catch` block be added when the tree is ticked so that exceptions are no longer fatal (I will be implementing this in an upcoming PR).

Issue #225, opened 2022-05-22
Add option for logging the execution of BTs [feature request]


sarcasticnature:
BT.CPP has a [useful logging feature](https://www.behaviortree.dev/tutorial_05_subtrees/) which allows you to save the behavior tree's execution and later play it back using Groot.  It would be nice to have this as an option when running a `bt_action_node`, especially for autonomous robots which cannot be actively monitored during operation.

I am currently working on this feature and will make a pull request soon

fmrico:
It sounds great!! :smiley:

PR #228, opened 2022-05-22
Add BT logging, Tree execution try/catch, and README

sarcasticnature:
Solves #225 and #226 and also adds a proper README to `plansys2_bt_actions` (which had a copy/pasted one from another package previously).


fmrico:
Thanks for implementing this and also for fixing the documentation.

Great work!!

PR #230, opened 2022-06-07
Expose lifecyclemanager timeout as a parameter, other misc fixes/features

### LINKED ISSUE #229: Uninitialized variable in ExecutorClient causes U.B.
This line: https://github.com/PlanSys2/ros2_planning_system/blob/c0d3550b6d915af168a9fbda9c4c93b1ecf40a4e/plansys2_executor/include/plansys2_executor/ExecutorClient.hpp#L66
In ExecutorClient leaves the `executing_plan_` variable uninitialized, as it isn't set in the class ctor.  This leads to undefined behavior in this line: https://github.com/PlanSys2/ros2_planning_system/blob/c0d3550b6d915af168a9fbda9c4c93b1ecf40a4e/plansys2_executor/src/plansys2_executor/ExecutorClient.cpp#L56
which may cause the ExecutorClient to falsely think that it is already executing a plan

Will fix in an upcoming PR

sarcasticnature:
Summary of changes
* fix #229 
* expose lifecycle manager timeout as a parameter in the bringup files
* the launch files now exit with a failure if they can't bring up the nodes
* added a ctor to ExecutorClient to enable unique node naming

### DIFF
--- a/plansys2_bringup/src/plansys2_node.cpp
+++ b/plansys2_bringup/src/plansys2_node.cpp
@@ -59,12 +59,14 @@ int main(int argc, char ** argv)
     exe.add_node(manager_node.second);
   }
 
-  std::shared_future<void> script = std::async(
+  std::shared_future<bool> script = std::async(
     std::launch::async,
     std::bind(plansys2::startup_script, manager_nodes));
   exe.spin_until_future_complete(script);
 
-  exe.spin();
+  if (script.get()) {
+    exe.spin();
+  }
 
   rclcpp::shutdown();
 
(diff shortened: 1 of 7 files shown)


sarcasticnature:
hmm, can't reproduce that exact error on my machine, but I have encountered an occasional bug where there are action performers available but they aren't found by the executor (it hasn't been consistent enough to track down yet).  May kick the runners again with a trivial change to see what happens

fmrico:
Hi @sarcasticnature 

Did you find the error? I will try to reproduce It too.

Thanks!!

fmrico:
Hi @sarcasticnature 

Nice work!! All tests are passing and everything seems to be working well. 

Thanks for your contribution!!

sarcasticnature:
> Hi @sarcasticnature
> 
> Did you find the error? I will try to reproduce It too.
> 
> Thanks!!

As I mentioned, the bug is very inconsistent and I can't replicate it intentionally.  If I ever discover the cause or a means to reliably replicate the bug I will file a GitHub issue

Issue #232, opened 2022-07-19
Check domain validity


devis12:
Good morning, I have a question related to the new implementation to check the domain validity doing at startup by the domain expert.
As before, you're running popf from the ros2 popf pkg with the loaded domain(s) and checking whether or not it gives an error. However, by running it on an empty problem file, it does not print anything if the domain is valid, nor it gives you any error (which it would if the domain is not valid). Nonetheless, you're checking if the ouput contains "Solution Found", which it never will, avoiding plansys2 to boot.
`
return result.find("Solution Found") != result.npos;`

I'm running on ROS2 Foxy and Mint 20.3 (should be Ubuntu Focal equivalent), but I don't think that's the issue, because I hadn't any before the pull that I performed this morning. Thank you for your support

devis12:
I would suggest to just check if there is any "error"/"Error" substring within the empty string that should be returnedas output by the planner.

fmrico:
Hi @devis12 

I think it could be a good fix. May you open a PR with your proposal?

Thanks!!

roveri-marco:
@fmrico I believe that the best solution is to check for "Solution found" as per your last fix to your modified version of popf or the output file is empty (as it was happening and it is expected in the standard version of popf).
cc: @jjzapf @sarcasticnature

jjzapf:
I closed my PR (#234) related to this issue because I thought it was solved. However, it appears that I accidentally tested the wrong branch. I will leave my PR closed as I do not think it was the right solution. I think we should continue the discussion here to determine the best path forward.

I concur with description of the issue provided by @devis12. In particular, I would note that this line creates a "mostly" empty problem file. I say mostly, because the file does contain the single line "(define (problem void) (:domain plansys2))"

https://github.com/PlanSys2/ros2_planning_system/blob/master/plansys2_popf_plan_solver/src/plansys2_popf_plan_solver/popf_plan_solver.cpp#L144

In the following line, we then run POPF using the domain specified at startup and the problem file defined above. When I run this command manually, it does not produce any output. However, @fmrico stated in #234 that he sees the following.

Constructing lookup tables:
Post filtering unreachable actions: 
Optimising makespan, lower bound is 0.000
;;;; Solution Found
; Time 0.00 

@fmrico, are you positive that you are seeing this output using the problem file specified above? I cannot seem to replicate this.

sarcasticnature:
[...]

PR #233, opened 2022-07-20
Domain validation fix

devis12:
With problem void, planner should return empty string not containing errors if the domain file is valid and cannot contain "Solution found"as it was check before

Edit: close issue #232 when accepted

### DIFF
--- a/plansys2_popf_plan_solver/src/plansys2_popf_plan_solver/popf_plan_solver.cpp
+++ b/plansys2_popf_plan_solver/src/plansys2_popf_plan_solver/popf_plan_solver.cpp
@@ -150,8 +150,9 @@ POPFPlanSolver::is_valid_domain(
 
   std::string result((std::istreambuf_iterator<char>(plan_file)),
     std::istreambuf_iterator<char>());
+  std::transform(result.begin(), result.end(), result.begin(), ::tolower);//result output plan to lower case
 
-  return result.find("Solution Found") != result.npos;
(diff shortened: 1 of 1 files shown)


plansys2_popf_plan_solver/src/plansys2_popf_plan_solver/popf_plan_solver.cpp
```diff
@@ -150,8 +150,9 @@ POPFPlanSolver::is_valid_domain(
 
   std::string result((std::istreambuf_iterator<char>(plan_file)),
     std::istreambuf_iterator<char>());
+  std::transform(result.begin(), result.end(), result.begin(), ::tolower);//result output plan to lower case
 
-  return result.find("Solution Found") != result.npos;
+  return result.find("error") == result.npos;//no "error" string in output from the planner -> domain valid
```
fmrico, line 155: I think this will not pass cpplint because spaces around `//`are needed (2 before and 1 after, I think), and also the line seems too long (is it <= 100). Please check it by running `colcon test`

jjzapf: I opened another MR for this same issue before realizing that this MR already existed. ... Sorry about that. My solution is a little different from this one, but essentially the same.

I'll leave my MR open in case you want to take a look. Feel free to decline, if you decide to go with this approach.

PR #235, opened 2022-08-10
[problem_expert] AddInstance successeeds if instance already exists

MostafaGomaa:
When adding an instance ( for example, through a service call to "problem_expert/add_problem_instance") , if the instance **already exists** we return false. This in turn causes the service to reply in failure with an error "instance not valid". This is both incorrect as well as inconsistent with how adding a predicate is handled. This PR fix that inconsistency and considers the service call successful if the instance is already there.

```cpp
ProblemExpertNode::add_problem_instance_service_callback( 
.....
   } else {                                                                    
       response->error_info = "Instance not valid";                              
     }
  .... 
 ```
 
 Changelog
 * FIX: [problem_expert] AddInstance successeeds if instance already exists

### DIFF
--- a/plansys2_problem_expert/src/plansys2_problem_expert/ProblemExpert.cpp
+++ b/plansys2_problem_expert/src/plansys2_problem_expert/ProblemExpert.cpp
@@ -41,14 +41,13 @@ ProblemExpert::ProblemExpert(std::shared_ptr<DomainExpert> & domain_expert)
 bool
 ProblemExpert::addInstance(const plansys2::Instance & instance)
 {
-  if (!isValidType(instance.type)) {
+  if (!isValidType(instance.type))
     return false;
-  } else if (existInstance(instance.name)) {
-    return false;
-  } else {
+
+  if (!existInstance(instance.name))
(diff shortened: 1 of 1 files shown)


MostafaGomaa:
Hey @fmrico ,

Is the CLI broken or is it the failure caused by my changes?
What I am getting is:
Screenshot_2022-08-10_18-44-57

I could not pin any of those failures to the changes.

Is there some doco on how can I run the tests locally?

Thanks and cheers,

fmrico:
It is a bug that affects other of my repos also. Let me try to fix CI...

fmrico:
Hi @MostafaGomaa 

I will merge this PR now, as you have been waiting for a long time for external reasons to you, and I don't want to bother you with more reviews after so long. **But** I will do a PR in 5 minutes, completing this PR with:

* It is necessary to check also if there is already a similar instance, but with another valid type.
* There are style problems that need to be resolved.
* If the behavior changes, it is necessary to change the tests

Thanks for your work!! You will see below the reference to the completing PR.

PR #236, opened 2022-08-24
Fixes for bt_actions and popf_plan_sover

sarcasticnature:
* solves #232
* solves issue in BTAction.cpp where loggers created during activation were not deleted during deactivation
* add try/catch to BT construction in BTAction.cpp
* (void) feedback in BtActionNode so that everyone's linters stop complaining

### DIFF
--- a/plansys2_bt_actions/include/plansys2_bt_actions/BTActionNode.hpp
+++ b/plansys2_bt_actions/include/plansys2_bt_actions/BTActionNode.hpp
@@ -120,6 +120,7 @@ public:
   virtual void on_feedback(
     const std::shared_ptr<const typename ActionT::Feedback> feedback)
   {
+    (void)feedback;
   }
 
   // Called upon successful completion of the action. A derived class can override this
--- a/plansys2_bt_actions/src/plansys2_bt_actions/BTAction.cpp
+++ b/plansys2_bt_actions/src/plansys2_bt_actions/BTAction.cpp
@@ -85,7 +85,15 @@ BTAction::on_cleanup(const rclcpp_lifecycle::State & previous_state)
 rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn
 BTAction::on_activate(const rclcpp_lifecycle::State & previous_state)
 {
-  tree_ = factory_.createTreeFromFile(bt_xml_file_, blackboard_);
+  try {
+    tree_ = factory_.createTreeFromFile(bt_xml_file_, blackboard_);
+  } catch (const std::exception & ex) {
+    RCLCPP_ERROR_STREAM(
+      get_logger(),
+      "Failed to create BT with exception: " << ex.what());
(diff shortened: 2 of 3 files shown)


sarcasticnature:
there are now three competing PRs.
https://xkcd.com/927/

sarcasticnature:
Linking to my comment on a previous PR: https://github.com/PlanSys2/ros2_planning_system/pull/239#issuecomment-1243208041

For reference:
> Alright, think I found the problem. The plansys2_bt_actions unit tests create and interact w/ a BT node outside of a tree and we were calling tick() instead of executeTick() to manually 'run' the node. The problem with the former is that it does not update the node's internal status (which the latter does), meaning that goals were being repeatedly sent to the action server because the node wasn't aware that it had already sent one.

The fix can be found in commit 836bc1b0aa933b6ea42786024294ee4c323d084d
The reason the fix works is because the conditional on the line below will always evaluate to `true` because the status isn't set at the end of each `tick()`

https://github.com/PlanSys2/ros2_planning_system/blob/d7f606e75bf584e643c6ecf27665700dedd75e78/plansys2_bt_actions/include/plansys2_bt_actions/BTActionNode.hpp#L150

calling `executeTick()` properly sets the status that the `status()` function gets

sarcasticnature:
Hmm, yet another failure but I think I may have this one figured out.  @fmrico did you make changes to `popf` in humble that were not backported to galactic?  I do all my development on galactic but switched over to a humble docker for today's troubleshooting (of the other CI failure) and have come to realize that the same domain/problem files cause `popf` to produce different output on the two distros

My fix is designed with the galactic behavior in mind, i.e. popf produces no output for a problem file with no goal (etc) and a domain that is valid, but I would obviously prefer the more affirmative humble output instead.  I hope that we can get the same behavior across both distros, if that is possible

fmrico:
Yes, I did a commit to popf (https://github.com/fmrico/popf/commit/bcc6cdf2d5161f8de56a5d5f1888b6848ace3821). Do you think these changes changed the behavior of popf? 🤔  I only fixed a null pointer bug and a compilation problem in newer gcc versions...

fmrico:
Or are you referring to the PlanSys2 popf plugin?

sarcasticnature:
I was referring to https://github.com/fmrico/popf.  I do in fact think fixing the null pointer bug changed the behavior of `popf`, if only because of the empirical evidence of running it before and after the change.

In addition, however, I just noticed that `popf` doesn't just produce no output for a problem w/o a goal (etc) on galactic, it actually fails with an error code (245, likely a segfault).  I am not familiar enough with `popf` to know if that is due to the bug you fixed, but it would appear so

If you could try running `popf` on both galactic and humble to add another data point to this behavior that would be appreciated.  The problem from the tests that causes an error is shown below, for reference:

```
(define (problem void) (:domain plansys2))
```

fmrico:
Hi @sarcasticnature 

I am accepting this PR, expecting to fix CI.

Thanks!!!

PR #239, opened 2022-09-08
Change CI to dockers

fmrico:
Hi

This PR is for fixing CI, which has been broken since some weeks: https://github.com/ros-tooling/action-ros-ci/issues/768

Best

### DIFF
--- a/.github/workflows/master.yaml
+++ b/.github/workflows/master.yaml
@@ -11,19 +11,17 @@ on:
 jobs:
   build-and-test:
     runs-on: ${{ matrix.os }}
+    container:
+      image: osrf/ros:humble-desktop
     strategy:
       matrix:
         os: [ubuntu-22.04]
       fail-fast: false
     steps:
-      - name: Install popf deps
-        run: sudo apt-get install libfl-dev
-      - name: Setup ROS 2
-        uses: ros-tooling/setup-ros@0.3.3
-        with:
-          required-ros-distributions: humble
+      - name: Install deps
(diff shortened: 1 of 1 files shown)


sarcasticnature:
hmmmm, it looks like the `on_feedback` test in `plansys2_bt_actions` is causing a failure (timing out) in the `master` CI... not quite sure why.  The amount of time it took the test to succeed is suspect even in this PR's (successful) test.  I don't think I've ever encountered the issue on my machine and am wondering if the test only has problems when operating on a slower runner.

Will investigate either tonight or this weekend.

fmrico:
Thanks, @sarcasticnature !! 

This issue blocks the PRs, and it's been a pain in the ass to fix it. I merged this PR because it passed the tests, but it failed again in the main...

sarcasticnature:
Alright, think I found the problem.  The `plansys2_bt_actions` unit tests create and interact w/ a BT node outside of a tree and we were calling `tick()` instead of `executeTick()` to manually 'run' the node.  The problem with the former is that it does not update the node's internal status (which the latter does), meaning that goals were being repeatedly sent to the action server because the node wasn't aware that it had already sent one.

If your system is fast (or you're just lucky) the test will likely eventually go through, but it's technically a race condition as is.  I have pushed an update to my PR (#236) which should resolve this issue

fmrico:
Thanks, @sarcasticnature!!! good catch ;)

As you push that commit in a merged PR, could you please make a PR only with the fix?

Thanks again

sarcasticnature:
I will try and get around to that tonight

PR #244, opened 2022-09-22
Horizontal bar chart from PDDL plan

jjzapf:
I created a simple Python script to construct a horizontal bar chart from a PDDL plan. This is especially useful for visualizing concurrency. I wasn't sure where to put this script, so I made a new package called plansys2_utils. Let me know if you have a better spot for it and I will move it.

Here is an example plan for the match domain that you can use to test the script.

    0.000: (light_match match1)  [8.000]
    0.001: (mend_fuse fuse1 match1)  [5.000]
    2.002: (light_match match2)  [8.000]
    5.002: (mend_fuse fuse2 match2)  [5.000]

Save this to a file named _match_plan.pddl_. To run the script, source the workspace and run the following command from the same directory containing the plan file.

    ros2 run plansys2_utils horizontal_bar_chart --ros-args -p plan_file:="match_plan.pddl"

You should see the horizontal bar chart displayed on the screen. The chart will also be saved to a PDF in the same directory containing the plan file.


jjzapf:
[match_plan.pdf](https://github.com/PlanSys2/ros2_planning_system/files/9630028/match_plan.pdf)

fmrico:
Hi @jjzapf 

I prefer not to increment the packages in this repo to avoid the complexity of producing the binary packages. 

I propose to create a new repo in this organization called `plansys2_utils` that will contain all the utils, a package for each one,  around PlanSys2. What do you think? We could even think of creating packages from some of them, but separately from the main distribution of PlanSys2.

If you agree, I think you can create it with your current role.

Francisco

jjzapf:
Hi @fmrico,

I'm good with your plan. However, I don't think I can create the repo. Could you create the repo?

Thanks!

fmrico:
Sorry for the delay, @jjzapf 

Created!!

Issue #245, opened 2022-10-06
Predicate 'notified' interpreted as 'not' node type


pac48:
I was getting a segmentation fault when trying to set a goal predicate using setGoal of the ProblemExpertClient class. 
```
problem_expert_->setGoal(plansys2::Goal("(and(robot_at pioneer door)(notified midnight_warning))"))
```
It appears that the regex [here](https://github.com/PlanSys2/ros2_planning_system/blob/99137b7f96aca3daf4c1667e9d87f67b854fbaee/plansys2_pddl_parser/src/plansys2_pddl_parser/Utils.cpp#L58) is not specific enough to known the difference between a not and notified.

My understanding is that valid syntax for not is the following:
```
(not (notified midnight_warning))
(not(notified midnight_warning))
```
So it should only be detected if not is followed by a space or a open parentheses.

PR #246, opened 2022-10-06
Fix regex for 'not', 'and', 'or' in parser utils

pac48:
This PR fixes predicates like 'notified' being detected as 'not', causing a crash. 
See this [issue](https://github.com/PlanSys2/ros2_planning_system/issues/245) for a more detailed explanation.


fmrico:
> Quick follow up question: should the regex for 'and' / 'or' also be changed in the `getNodeType` method?

Yes!!! Thanks @pac48

fmrico:
Thanks @pac48 

Merged :rocket:

PR #248, opened 2022-10-29
Add pre-commit for automatic formatting

pac48:
This PR adds a pre-commit config to run clang formatting. I checked the [CONTRIBUTING.md](https://github.com/PlanSys2/ros2_planning_system/blob/061a37191d6be5b9f813693f230a25e7aa573f0c/CONTRIBUTING.md) and didn't see any format requirement. For anyone not familiar with pre-commit, here is a helpful tutorial:  [Check formatting locally and on GitHub Actions with pre-commit](https://www.youtube.com/watch?v=4K6GZWFQs3A)

In short, pre-commit can be installed via: `pip install pre-commit`
and ran using: `pre-commit run --all`

The formatting is setup to run as part of the CI.


fmrico:
Hi @pac48 

I need a little more context in this PR. What are the benefits of using clang formatting instead of the ROS 2 standard?

Best
Francisco

pac48:
> Hi @pac48
> 
> I need a little more context in this PR. What are the benefits of using clang formatting instead of the ROS 2 standard?
> 
> Best Francisco

Sorry for not elaborating, let me explain my issue. I opened a PR [Add Support for Contingency Planning](https://github.com/PlanSys2/ros2_planning_system/pull/249), which has a huge diff because I used a different format. That is no problem, the code can be reformatted, so I thought adding clang automatic formatting as part of CI would be nice.      

One advantage of the clang-format over the ament based tools is the de facto standard in C++. It is a consistent and powerful tool for formatting. The real downside to clang-format is that it is kind of annoying to run locally. But that is where pre-commit comes in. It is a python script that automates running tools like clang-format locally for developers and can also be used in CI to verify the formatting stays consistent. With pre-commit, you can also verify the formatting of python files, spell check, check YAML formatting, etc. The list goes on. When applying to the this repo, there were actually quite a few changes recommended by the python formater "black". There were also many spelling mistakes that were caught suggestions recommended.   

The biggest problem with clang-format is that it can't be applied to existing ROS 2 code bases because there are some small formatting exceptions it does not allow. See [here](https://github.com/ament/ament_lint/issues/146) for a discussion.   

I don't have a strong opinion on which formater should be sued, although it should be within the ROS code style guidelines of course. With that said, here is a branch from master after running `ament_uncrustify --reformat`
[run-ament-reformat](https://github.com/pac48/ros2_planning_system/tree/run-ament-reformat). You can see there is a huge diff still. I think having some automatic formatting will save headaches in the future. 

Is there some other tool (besides clang-format and ament_uncrustify) that you recommend?

fmrico:
I have not forgotten this PR. I need time to think deeply about this.

PR #249, opened 2022-10-30
Add Support for Contingency Planning

pac48:
## Contingency planning
This PR adds support for contingency planning with the contingent FF planner. The planner plugin can be found here: https://github.com/pac48/plansys2_cff_plan_solver. I made a formatting [PR](https://github.com/PlanSys2/ros2_planning_system/pull/248) that will need to be reviewed before this one because this PR is rebased on it (the diff is currently showing all of the formatting changes). I tried to keep the changes as possible, but there are many tests and added pddl files that make the change very large.

## Contingency planning demo
We are interested in applying for at home robot care giving. One protocol the robot must deliver is ensuring medicine is taken. Since it is not not ahead of time if the person will comply with the robot's request, we need to plan all contingencies. Hence, we are implementing pddl based contingency planners for our robot decision making.

In this scenario, the robot does not know where the person is in the house, but knows one of a few places the person might be. After finding the person, the robot makes a request to follow it to the kitchen to take the medicine. In the case the person does not follow the robot, the robot will try a second time and then, finally, call a human caregiver for help. If the person follows the robot to the kitchen, the person is then prompted to take the medicine. Again, they may or may not take it, and the robot will try two prompts before calling a human caregiver. The behavior we generated using continent FF is show here: bt We have a video demonstration in simulation of two different outcomes: [video](https://universitysystemnh-my.sharepoint.com/:v:/g/personal/pac48_usnh_edu/EXxUE8B22GNKgoKxoNtihbkBwQik6LMAgCCl2gTpRCI5DA?e=I3ZKNH)

## Known issues
1. Previously, domains were validated using the default popf solver, but this needs to chnage becuase popf does not support the full pddl syntax.  
2. The RQT plugins currently do not show the problem constraints, e.g. oneof or unknown.
3. The plansys2_terminal does not support oneof or unknown   
4. The contingency behavior tree plugin does not currently publish its dot graph


roveri-marco:
This work is interesting. However, it is not clear how time and concurrent execution of actions are handled.
Moreover, it seems to break the other packages. In order to be integrated and merged the full set of packages need to be modified (terminal, domain, problem expert, ...) in a way that it should be backwards compatible. If one wants to use contingent planning then it selects this modality and all will be configured to work accordingly, otherwise, it will work as before (i.e. temporal planning).

pac48:
> This work is interesting. However, it is not clear how time and concurrent execution of actions are handled. Moreover, it seems to break the other packages. In order to be integrated and merged the full set of packages need to be modified (terminal, domain, problem expert, ...) in a way that it should be backwards compatible. If one wants to use contingent planning then it selects this modality and all will be configured to work accordingly, otherwise, it will work as before (i.e. temporal planning).

Thank you for your feedback. Its been some time since I worked on this project, so I may some time before I can address substantial changes. That being said, I added a new executor node in this PR that can be enabled/disabled with a ROS parameter. Ideally, it should be a simple launch argument. Since the contingent executor is independent from the parallel executor, I don't see how it can cause issues. I can investigate these issues. 

Since this PR depends on this [one](https://github.com/PlanSys2/ros2_planning_system/pull/248) my main focus getting that approved or I can try to rebase my changes without the formatting.

PR #250, opened 2022-11-23
Time triggered exec v4

jjzapf:
I've implemented the propagation step of the get_tree function in the STN BT Builder plugin using the Floyd-Warshall algorithm. This step is necessary to propagate the time constraints through the edges of the graph. I've also added the match-cellar example to ros_planning_system_exmaples (see match-example branch) to help test this. Without these changes, the example will fail due to the fact that the 2nd match is lit immediately rather than 2 seconds after the first.

To run this example ...

ros2 launch plansys2_match_example plansys2_match_example_launch.py

ros2 run plansys2_terminal plansys2_terminal

set instance m1 match
set instance m2 match
set instance f1 fuse
set instance f2 fuse
set predicate (unused m1)
set predicate (unused m2)
set predicate (handfree)
set goal (and (mended f1) (mended f2))
get plan
run


fmrico:
Excellent!! I will try to test it in my computer this evening

plansys2_domain_expert/src/plansys2_domain_expert/DomainExpertNode.cpp
```diff
@@ -117,7 +117,7 @@ DomainExpertNode::on_configure(const rclcpp_lifecycle::State & state)
   domain_expert_ = std::make_shared<DomainExpert>(domain_str);
 
   bool check_valid = planner->is_valid_domain(domain_expert_->getDomain(), get_namespace());
-  if (!check_valid) {
+  if (check_valid) {
```
fmrico, line 120: I think that the correct semantic should be `if not valid then print error`

plansys2_executor/include/plansys2_executor/ActionExecutor.hpp
```diff
@@ -66,6 +66,7 @@ class ActionExecutor
   std::vector<std::string> get_action_params() const {return action_params_;}
   plansys2_msgs::msg::ActionExecution last_msg;
 
+  rclcpp::Time get_current_time() const {return node_->now();}
```
fmrico, line 69: Is not this function redundant, if you have a Node?

plansys2_executor/include/plansys2_executor/bt_builder_plugins/stn_bt_builder.hpp
```diff
@@ -15,6 +15,8 @@
 #ifndef PLANSYS2_EXECUTOR__BT_BUILDER_PLUGINS__STN_BT_BUILDER_HPP_
 #define PLANSYS2_EXECUTOR__BT_BUILDER_PLUGINS__STN_BT_BUILDER_HPP_
 
+#include <eigen3/Eigen/Dense>
```
fmrico, line 18: Is this dependency included in the `package.xml`?

fmrico:
Hi @jjzapf 

It blocks to me. May you check it?

image

plansys2_domain_expert/src/plansys2_domain_expert/DomainExpertNode.cpp
jjzapf: Opps. I reverted this change. ... For some reason I still need to do this on my Foxy build, but I assume it is something wrong on my end. ... Previously I was able to resolve this by switching to the ros2 branch of popf. However, that does not seem to work any longer. I will try upgrading to Humble soon.

plansys2_executor/include/plansys2_executor/ActionExecutor.hpp
jjzapf: Yes. Reverting this. See cf4a474

plansys2_executor/include/plansys2_executor/bt_builder_plugins/stn_bt_builder.hpp
jjzapf: I added the Eigen3 dependency to CMakeLists.txt and package.xml. See 6119d12

jjzapf:
@fmrico, the BT you attached to your comment does not match the BT that I get. Could you double check to see if you are on the right branch? The branch should be time-triggered-exec-v4 for ros2_planning_system and match-example for ros2_planning_system_examples.

fmrico:
Hi @jjzapf 

I continue with the same blocking problem after changing to the `match-example` branch

image
[bt.zip](https://github.com/PlanSys2/ros2_planning_system/files/10097991/bt.zip)

PR #272, opened 2023-09-16
Add ability to specify output folder in POPF planner

sea-bass:
This PR allows you to specify a `<plugin_name>.output_dir` argument to specify an alternative output folder to the default temp folder `/tmp`.

We're motivated by this because we may want to specifically spit out the domain/problem PDDL files and resulting plan to another folder.

I also cleaned up a bit of the logic to better use the `std::filesystem` library and reduce duplication.

### DIFF
--- a/plansys2_domain_expert/src/plansys2_domain_expert/DomainExpertNode.cpp
+++ b/plansys2_domain_expert/src/plansys2_domain_expert/DomainExpertNode.cpp
@@ -114,6 +114,7 @@ DomainExpertNode::on_configure(const rclcpp_lifecycle::State & state)
     std::istreambuf_iterator<char>());
 
   auto planner = std::make_shared<plansys2::POPFPlanSolver>();
+  planner->configure(shared_from_this(), "POPF");
   domain_expert_ = std::make_shared<DomainExpert>(domain_str);
 
   bool check_valid = planner->is_valid_domain(domain_expert_->getDomain(), get_namespace());
--- a/plansys2_domain_expert/src/plansys2_domain_expert/DomainExpertNode.cpp
+++ b/plansys2_domain_expert/src/plansys2_domain_expert/DomainExpertNode.cpp
@@ -108,27 +108,20 @@ DomainExpertNode::on_configure(const rclcpp_lifecycle::State & state)
 
   auto model_files = tokenize(model_file, ":");
 
-  std::ifstream domain_ifs(model_files[0]);
-  std::string domain_str((
-      std::istreambuf_iterator<char>(domain_ifs)),
-    std::istreambuf_iterator<char>());
-
-  auto planner = std::make_shared<plansys2::POPFPlanSolver>();
+  auto planner = std::make_unique<plansys2::POPFPlanSolver>();
   planner->configure(shared_from_this(), "POPF");
-  domain_expert_ = std::make_shared<DomainExpert>(domain_str);
 
-  bool check_valid = planner->is_valid_domain(domain_expert_->getDomain(), get_namespace());
-  if (!check_valid) {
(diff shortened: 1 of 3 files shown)


plansys2_domain_expert/src/plansys2_domain_expert/DomainExpertNode.cpp
```diff
@@ -114,6 +114,7 @@ DomainExpertNode::on_configure(const rclcpp_lifecycle::State & state)
     std::istreambuf_iterator<char>());
 
   auto planner = std::make_shared<plansys2::POPFPlanSolver>();
+  planner->configure(shared_from_this(), "POPF");
```
sea-bass, line 117: Caught this bug with these changes -- we were relying on an unconfigured solver node here.

adlarkin:
The documentation in https://plansys2.github.io/design/index.html#planner will also need to be updated, it's currently referring to the `/tmp` folder.

plansys2_popf_plan_solver/src/plansys2_popf_plan_solver/popf_plan_solver.cpp
```diff
@@ -50,37 +55,46 @@ POPFPlanSolver::getPlan(
...
   problem_out << problem;
   problem_out.close();
 
-  int status = system(
-    ("ros2 run popf popf " +
-    lc_node_->get_parameter(parameter_name_).value_to_string() +
-    " /tmp/" + node_namespace + "/domain.pddl /tmp/" + node_namespace +
-    "/problem.pddl > /tmp/" + node_namespace + "/plan").c_str());
+  const auto plan_file_path = output_dir / std::filesystem::path("plan");
+  const auto args = lc_node_->get_parameter(arguments_parameter_name_).value_to_string();
```
fmrico, line 86: I have seen that references to `node_namespace` in your PR have been removed. Are your changes ready to use namespaces? We must support having different PlanSys2 instances, one in each namespace for not collide

fmrico:
> The documentation in https://plansys2.github.io/design/index.html#planner will also need to be updated, it's currently referring to the `/tmp` folder.

Yes, to accept this PR, we need to do another PR in https://github.com/PlanSys2/PlanSys2.github.io to adapt to these changes. Can you do it, @sea-bass ?

plansys2_domain_expert/src/plansys2_domain_expert/DomainExpertNode.cpp
fmrico: good catch :+1:

plansys2_popf_plan_solver/src/plansys2_popf_plan_solver/popf_plan_solver.cpp
sea-bass: Not at all! The namespaces continue to be added in this loop: https://github.com/PlanSys2/ros2_planning_system/blob/2d8ff58b6188a145ad86e2c4e8d510e29534dbd1/plansys2_popf_plan_solver/src/plansys2_popf_plan_solver/popf_plan_solver.cpp#L147-L153

In my tests, the results are definitely written to `<output_dir>/<node_namespace>` (or `<temp_dir>/<node_namespace>` by default)

fmrico:
Ok. Then, this PR is ready to merge. I will merge it at the same time of the PR in the documentation

Thanks, @sea-bass !!!!

sea-bass:
> Ok. Then, this PR is ready to merge. I will merge it at the same time of the PR in the documentation
> 
> Thanks, @sea-bass !!!!

Done! Refer to https://github.com/PlanSys2/PlanSys2.github.io/pull/13

fmrico:
Merged!!!M 🚀 

Thanks @sea-bass

Issue #167, opened 2021-12-10
cannot use plan with numeric-fluents (which is allowed in TFD planner)


edhage:
when I use TDF-planner still I get an error "Error parsing PDDL: numeric-fluents does not name a known token" in my domain-expert-node, and also the problem-expert-node.

The reason for this is that 
```auto planner = std::make_shared<plansys2::POPFPlanSolver>();``` is loaded in the domain expert node to check the model, even when I specify the TFD-planner. 

This should be changed. The domain expert should check with the TFD-planner and allow numeric-fluents in that case.

fmrico:
Hi @edhage 

The decision to use POPF to check the model was made because it was the only solver that was distributed together with PlanSys2. Of course, what you propose seems totally reasonable.

If you want to work on a PR to solve it, I will tell you how to approach it:
1. Add to `plansys2::PlanSolverBase` a pure virtual `check_domain` method, as in `plansys2 :: POPFPlanSolver`
2. Implement this method in `plansys2::TFDPlanSolver`
3. In `plansys2_params.yaml` there should be a parameter for the domain expert specifying the solver, and depending on it, one or the other will be instantiated in `plansys2::DomainExpertNode::on_configure`

Another option would be to use something generic like the one in the package `plansys2_pddl_parser`, but it would be necessary to evaluate if it is sufficiently general.

In any case, I'll be happy to help you and discuss the details in this issue or in a PR when you get started.

Best!!

roveri-marco:
The PR #204 GH-204 should also be able to fix this issue.

sea-bass:
Maybe `PlanSolverBase` can expose a ROS service that wraps around this plan validation function.

This would provide a way to not have to instantiate a new planner in the domain expert, but rather wait and call for the service from the planner.

... and with some option (maybe a boolean parameter), we could toggle between the behavior of waiting for a planner to validate the domain vs. creating a POPF instance and validating it locally.

sea-bass:
Update, I've put up 2 PRs that should allow for this:

* https://github.com/PlanSys2/ros2_planning_system/pull/275
* https://github.com/PlanSys2/plansys2_tfd_plan_solver/pull/12

Looking forward to getting these fixed up and in a mergeable state!

PR #275, opened 2023-09-23
Add option to use planner node to validate domain in domain expert

sea-bass:
This PR adds an option for the domain expert to use a planner node directly to validate the domain.

This is enabled in the domain expert through a `validate_using_planner_node` ROS parameter, which defaults to `false` to maintain compatibility. If false, a POPF instance will be created as before to validate the domain. Otherwise, a ROS service will directly call out to the planner node's planner to do the domain validation.

As such, this enforces that plan solvers now implement an `isDomainValid()` method, which is wrapped in a `planner/validate_domain` service by the `PlannerNode`.

The idea here is that if using a solver that is not POPF, we don't have to require the domain to be valid according to POPF. See e.g. this issue: https://github.com/PlanSys2/ros2_planning_system/issues/167

### DIFF
--- a/plansys2_domain_expert/src/plansys2_domain_expert/DomainExpertNode.cpp
+++ b/plansys2_domain_expert/src/plansys2_domain_expert/DomainExpertNode.cpp
@@ -19,7 +19,6 @@
 #include <vector>
 
 #include "plansys2_core/Utils.hpp"
-#include "plansys2_popf_plan_solver/popf_plan_solver.hpp"
 
 #include "lifecycle_msgs/msg/state.hpp"
 
@@ -30,6 +29,7 @@ DomainExpertNode::DomainExpertNode()
 : rclcpp_lifecycle::LifecycleNode("domain_expert")
 {
   declare_parameter("model_file", "");
+  declare_parameter("validate_using_planner_node", false);
 
   get_name_service_ = create_service<plansys2_msgs::srv::GetDomainName>(
     "domain_expert/get_domain_name",
@@ -94,6 +94,8 @@ DomainExpertNode::DomainExpertNode()
       &DomainExpertNode::get_domain_service_callback,
       this, std::placeholders::_1, std::placeholders::_2,
       std::placeholders::_3));
+
+  validate_domain_client_ = create_client<plansys2_msgs::srv::ValidateDomain>("planner/validate_domain");
 }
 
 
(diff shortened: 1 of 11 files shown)


plansys2_domain_expert/src/plansys2_domain_expert/DomainExpertNode.cpp
```diff
@@ -123,7 +135,20 @@ DomainExpertNode::on_configure(const rclcpp_lifecycle::State & state)
...
-    bool check_valid = planner->is_valid_domain(domain_expert_->getDomain(), get_namespace());
+    bool check_valid = true;
+    if (validate_using_planner_node) {
+      auto request = std::make_shared<plansys2_msgs::srv::ValidateDomain::Request>();
+      request->domain = domain_expert_->getDomain();
+      auto future_result = validate_domain_client_->async_send_request(std::move(request));
+      if (future_result.wait_for(std::chrono::seconds(1)) != std::future_status::ready) {
+        RCLCPP_ERROR(get_logger(), "Timed out waiting for service: %s",
+          validate_domain_client_->get_service_name());
+        return CallbackReturnT::FAILURE;
```
sea-bass, line 146: I'm not sure why this always times out... must be something weird about rclcpp

sea-bass:
@fmrico If you think this approach looks okay, I'll make sure to update the TFD solver to also implement `isValidDomain()` and will document the changes in the docs.

plansys2_planner/src/plansys2_planner/PlannerNode.cpp
```diff
@@ -156,4 +162,18 @@ PlannerNode::get_plan_service_callback(
...
+  const std::shared_ptr<rmw_request_id_t> request_header,
+  const std::shared_ptr<plansys2_msgs::srv::ValidateDomain::Request> request,
+  const std::shared_ptr<plansys2_msgs::srv::ValidateDomain::Response> response)
+{
+  response->success = solvers_.begin()->second->isDomainValid(
+    request->domain, get_namespace());
+
+  if (!response->success) {
+    response->error_info = "Domain is not valid";
+  }
```
sea-bass, line 176: Another option is for `isValidDomain()` to return a string error message that can also be tacked on to the service response...

plansys2_planner/src/plansys2_planner/PlannerNode.cpp
```diff
@@ -90,6 +83,20 @@ PlannerNode::on_configure(const rclcpp_lifecycle::State & state)
...
+      &PlannerNode::get_plan_service_callback,
+      this, std::placeholders::_1, std::placeholders::_2,
+      std::placeholders::_3));
+
+  validate_domain_service_ = create_service<plansys2_msgs::srv::ValidateDomain>(
+    "planner/validate_domain",
+    std::bind(
+      &PlannerNode::validate_domain_service_callback,
+      this, std::placeholders::_1, std::placeholders::_2,
+      std::placeholders::_3));
```
sea-bass, line 98: Needed to move the service instantiations here since I ran into timing issues where another node could call these services before the planner was actually instantiated. I think this is good to fix regardless.

plansys2_domain_expert/src/plansys2_domain_expert/DomainExpertNode.cpp
```diff
@@ -104,12 +106,26 @@ CallbackReturnT
...
+  const std::string model_file = get_parameter("model_file").get_value<std::string>();
+  const bool validate_using_planner_node =
+    get_parameter("validate_using_planner_node").get_value<bool>();
 
   auto model_files = tokenize(model_file, ":");
 
-  auto planner = std::make_unique<plansys2::POPFPlanSolver>();
-  planner->configure(shared_from_this(), "POPF");
+  if (validate_using_planner_node) {
+    // validate_domain_node_ = rclcpp::Node::make_shared("temporary_domain_validation_node");
```
sea-bass, line 116:
```suggestion
```

fmrico:
Hi @sea-bass 

This feature is cool!! This resolves some problems that you point to. 

Merging!! :rocket: 

Thanks!!!

PR #281, opened 2023-11-16
Make `ProblemExpert` case-insensitive to Instance types

### LINKED ISSUE #280: `ProblemExpert::addInstance` is case-sensitive, but `ProblemExpert::addPredicate` is not
Consider the following code:

```c++
const auto domain_test_str = R"(
(define (domain simple)
(:requirements :strips :typing :adl :fluents)

(:types
    Robot
    Location
    Object
)

;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
; Predicates, Functions, and Actions can be defined below ... ;
;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;

)
)";

auto domain_expert = std::make_shared<plansys2::DomainExpert>(domain_test_str);

plansys2::ProblemExpert problem_expert(domain_expert);

problem_expert.addInstance(plansys2::Instance("rob", "Robot"));
problem_expert.addInstance(plansys2::Instance("table", "Location"));
problem_expert.addInstance(plansys2::Instance("apple", "Object"));
```

The `problem_expert.addInstance` calls fail, because [the domain is converted to all lowercase](https://github.com/PlanSys2/ros2_planning_system/blob/master/plansys2_domain_expert/src/plansys2_domain_expert/DomainReader.cpp#L47), but [`ProblemExpert::addInstance` does not convert the instance type to lowercase](https://github.com/PlanSys2/ros2_planning_system/blob/master/plansys2_problem_expert/src/plansys2_problem_expert/ProblemExpert.cpp#L402) in the method implementation.

If the `DomainExpert` is converting everything to lowercase, then other API calls that use the domain should also convert inputs to lowercase. This is being done when calling [`ProblemExpert::addPredicate`](https://github.com/PlanSys2/ros2_planning_system/blob/master/plansys2_problem_expert/src/plansys2_problem_expert/ProblemExpert.cpp#L461), and `ProblemExpert::addInstance` should follow the same pattern.

adlarkin:
Closes #280 

My interpretation here is that Instance _names_ should be case-sensitive, but instance _types_ shouldn't be.

### DIFF
--- a/plansys2_problem_expert/src/plansys2_problem_expert/ProblemExpert.cpp
+++ b/plansys2_problem_expert/src/plansys2_problem_expert/ProblemExpert.cpp
@@ -41,19 +41,29 @@ ProblemExpert::ProblemExpert(std::shared_ptr<DomainExpert> & domain_expert)
 bool
 ProblemExpert::addInstance(const plansys2::Instance & instance)
 {
-  if (!isValidType(instance.type)) {
+  plansys2::Instance lowercase_instance = instance;
+  std::transform(
+    lowercase_instance.type.begin(), lowercase_instance.type.end(), lowercase_instance.type.begin(),
+    [](unsigned char c) {return std::tolower(c);});
+  for (auto i = 0; i < lowercase_instance.sub_types.size(); ++i) {
+    std::transform(
+      lowercase_instance.sub_types[i].begin(), lowercase_instance.sub_types[i].end(),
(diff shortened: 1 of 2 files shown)


fmrico:
HI @adlarkin 

LGTM, and the tests are passing: **merging**!!! :rocket: 

Thanks!!

Issue #169, opened 2021-12-29
Missing Functionality for "Imply" language feature


NmBoyd:
If I am understanding everything correctly. It appears that there is currently no support for the `imply` language feature as defined in PDDL 2.1. Which would be quite helpful for some of the task planning problems I am looking to solve. Notably for  action preconditions specified like: `(forall (....) (imply (antecedent_predicate) (consequent_predicate))))`.

When could I perhaps expect support for this feature? Or go about implementing it myself? Help would be appreciated. 

Currently this prevents the creation of desired actions. From what I understand, I would have to create a new condition in the Domain [here](https://github.com/IntelligentRoboticsLabs/ros2_planning_system/blob/47c4e6f1cccf6e6a5683a75c44ba217f73c8f55a/plansys2_pddl_parser/include/plansys2_pddl_parser/Domain.h) based on a class that contains the parsing logic.

fmrico:
Dear @NmBoyd 

I've never used `imply`, and any other contributor never rasked for it before. You'll need to implement it by yourself.

In a quick search, `pddl::parse` in package `plansys2_pddl_parser` does not seem to support `imply so it would be necessary to add support there. 

PDDL plan solver program should support it, that is not clear to me. You can also use a plan solver that supports it. PlanSys2 currently POPF (default), Optics and TFD. If you want to use another, you'll need to create a plugin like the one in `plansys2_popf_plan_solver`.

Best
Francisco

NmBoyd:
Sorry for the late reply on this. Thanks for the pointers on this. I do see a direction moving forward with doing this then. 
I've found a temporary work around for now to solve my problem, but I'll try to circle back around to looking at an implementation to this at a later date. Can close this issue for now.

sea-bass:
FYI @NmBoyd -- I know it's almost 2 years later but here it goes: https://github.com/PlanSys2/ros2_planning_system/pull/283

PR #283, opened 2023-11-23
Add Imply to PDDL domain parsing

sea-bass:
I was running PlanSys2 with a PDDL domain that had an `imply` in it, and noticed we did not support it.

So, now we do! Thankfully it was basically copypasta from the `forall` implementation that was already there... and it works :)

Closes https://github.com/PlanSys2/ros2_planning_system/issues/169

### DIFF
--- a/plansys2_pddl_parser/include/plansys2_pddl_parser/Domain.h
+++ b/plansys2_pddl_parser/include/plansys2_pddl_parser/Domain.h
@@ -314,19 +314,6 @@ public:
 		actions.insert( a );
 	}
 
-	// void parseImply( Stringreader & f ) {
-	// 	if ( !preds.size() ) {
-	// 		std::cout << "Predicates needed before defining derived predicates\n";
-	// 		exit(1);
-	// 	}
-
-	// 	f.next();
-	// 	Imply * d = new Imply;
-	// 	d->parse( f, types[0]->constants, *this );
-
-	// 	if ( DOMAIN_DEBUG ) std::cout << d << "\n";
-	// 	derived.insert( d );
-	// }
 
 	// Return a copy of the type structure, with newly allocated types
 	// This will also copy all constants and objects!
--- a/plansys2_pddl_parser/include/plansys2_pddl_parser/Domain.h
+++ b/plansys2_pddl_parser/include/plansys2_pddl_parser/Domain.h
@@ -9,6 +9,7 @@
   #include "plansys2_pddl_parser/Forall.h"
   #include "plansys2_pddl_parser/Function.h"
   #include "plansys2_pddl_parser/GroundFunc.h"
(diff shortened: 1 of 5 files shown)


plansys2_pddl_parser/include/plansys2_pddl_parser/Domain.h
```diff
@@ -313,6 +314,19 @@ class Domain {
...
+	// 		exit(1);
+	// 	}
+
+	// 	f.next();
+	// 	Imply * d = new Imply;
+	// 	d->parse( f, types[0]->constants, *this );
+
+	// 	if ( DOMAIN_DEBUG ) std::cout << d << "\n";
+	// 	derived.insert( d );
+	// }
```
sea-bass, line 329:
```suggestion
```

fmrico:
Hi @sea-bass 

Could you add a simple test to verify that it is correctly parsed? Tests in this package are simple.

Thanks!!

sea-bass:
> Hi @sea-bass
> 
> Could you add a simple test to verify that it is correctly parsed? Tests in this package are simple.
> 
> Thanks!!

Done! Added both `forall` (which wasn't there) and `imply` to the parser test domain. Let me know if you had something else in mind.

fmrico:
Thanks!!! Merging :rocket:

eshan-savla:
Hi @sea-bass,

First of all, thank you so much for your contribution! I don't mean to undermine your efforts with this, but I was going over your changes as a part of this issue #344 and something stood out to me. 

```C++
void Imply::parse( Stringreader & f, TokenStruct< std::string > & ts, Domain & d ) {
1.	f.next();
2.	f.assert_token( "(" );

3.	TokenStruct< std::string > fs = f.parseTypedList( true, d.types );
4.	params = d.convertTypes( fs.types );

5.	TokenStruct< std::string > fstruct( ts );
6.	fstruct.append( fs );

7.	f.next();
8.	f.assert_token( "(" );
9.	if ( f.getChar() != ')' ) {
10.		cond = d.createCondition( f );
11.		cond->parse( f, fstruct, d );
12.	}
13.	else ++f.c;

14.	f.next();
15.	f.assert_token( ")" );
}
```
I haven't been working with this repo and pddl for so long, so I might have a few things wrong, but I have a question. Lines 3 - 4 make sense in the case of _forall_, where the first set of tokens are always typed lists, however in the case of _imply_, the first set of tokens aren't typed lists, rather predicates. Considering this, would it not make more sense to somehow reuse lines 9-12? Also, the second predicate is being set as the condition, whereas in truth, shouldn't it be an evaluation of both predicates of imply or just the first one? Maybe saving both as two seperate conditions?

I might also be completely off base in my line of thought, so perhaps someone else could also weigh in and let me know if my thinking is on the right track? @fmrico, if you have the time, I would greatly value your thoughts on this.

PR #284, opened 2023-11-25
Lintering and fix after stn bugs

fmrico:
Hi

I have detected some bugs that made fail some executions. This PR is a WIP to fix them.

Best


fmrico:
I have detected bugs in the BT nodes that compose the execution tree in Executor. These bugs are hard to fix and are only related to the tree created using STN. Until these bugs are fixed, I decided to restore a pre-STN version of these files. As soon as the STN nodes are fixed and tested, we can include STN-related code again. I have also disabled the option to use the STN BT builder, displaying a warning message.

PR #289, opened 2024-02-06
Adding action duration in dotgraph.

xydesa:


xydesa:
Here's an example dotgraph from the Simple BT Builder:
ugv-demo-simple-builder

xydesa:
Here's an example dotgraph from the STN BT Builder:
ugv-demo-stn-builder

fmrico:
Hi @xydesa 

Thanks for your PR. As the tests are passing, merging!!! 🚀

PR #296, opened 2024-03-01
Removed some small references still related to the old version of BehaviorTree.CPP 3.8

SamueleSandrini:
I removed references/refusals still related to the old version of _Behavior Tree V3.8_. In particular:

- In the `plansys2_bt_actions` package in the `CmakeLists` there was a repetition of `include_directories(include) `and in the `package.xml` there was still the dependency to `libzmq3-dev` related to `zmq_publisher` for BT V3.8 communication with Groot which was removed from the latest PR.
- I added the `BTCPP_format="4"` in two BT xmls where it did not appear.
- In `plansys2_executor` there was still the old reference to `BT::NodeConfiguration` now `BT::NodeConfig` in `BT 4.x` naming convention.

### DIFF
--- a/plansys2_bt_actions/CMakeLists.txt
+++ b/plansys2_bt_actions/CMakeLists.txt
@@ -68,6 +68,5 @@ endif()
 
 ament_export_include_directories(include)
 ament_export_dependencies(${dependencies})
-include_directories(include)
 
 ament_package()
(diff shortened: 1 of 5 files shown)


fmrico:
Thanks @SamueleSandrini 

Merging!!! :rocket:

PR #299, opened 2024-03-01
[executor] Send Cancelation response to ActionClient

MostafaGomaa:
Currently, When the ExecutorNode (i.e., ActionServer) receives cancellation of `execute_plan` goal, it halts the execution of the behavior tree. It is also desired that action-servers send a cancellation response. This allows the action-client to trigger on_cancellation hooks to finalize their goal.


fmrico:
Merging!!! :rocket: 

Tkanks

PR #300, opened 2024-03-02
[executor] fail execute_plan action if initial conditions changed

MostafaGomaa:
We terminate the program is many scenarios instead of just returning an explicit value. We currently terminate if the initial conditions change before the plan is started. This is prone to happen in the real world.

This PR sends an explicit result with a failure result to "execute_plan" goals for which the initial requirements change after planning and before execution.

* FEATURE:  [ComputeBT] Response with failure when unable to build tree
* FIX:  [ComputeBT] Joins thread when unable to build tree
* FEATURE:  [ExecutorNode] Send Result with failure when unable to build tree and publish an empty plan
* FIX: [simple_bt_builder] Don't terminate the program when requirements change. instead, return explicit values


fmrico:
Merging!!! :rocket: 

Thanks @MostafaGomaa

PR #302, opened 2024-03-04
Fix POPF error with empty problem

Butakus:
We found that, in some cases, POPF segfaults when the problem definition does not include the :objects, :init and :goal sections. That caused the domain_expert to report a "PDDL syntax error" message and crash.
The issue happens when checking if a domain is valid, because an empty problem is auto-generated and passed to POPF.

This PR should fix this issue.


fmrico:
Hi @Butakus 

CI is failing. It is strange because I know this resolves the "PDDL syntax error" error, but there is something strange with popf.  We need to dig more into this problem. 

I will close this PR now, but it is still helpful for documenting this and user could patch this problem.

Best

PR #303, opened 2024-03-05
[executor] ActionExecutor, finialze on cancel, its cancel and forget

MostafaGomaa:
Not sure yet if this is required Since when we cancel in a fire-and-forget fashion it might be needed to finalize and deactivate the publisher upon cancellation.


fmrico:
I think something is failing here.... :-/

fmrico:
Can you merge with `rolling`to trigger fixed CI?

Issue #243, opened 2022-09-19
Searching for a solution timeout shall be configurable


roveri-marco:
Currently ```plansys2_planner/src/plansys2_planner/PlannerClient.cpp``` has an hardcoded value of 15 seconds to find a solution plan. This might be ok for small examples. However, if the problem is a bit more complex 15 seconds might not be enough.

It would be good to have the possibility to programmatically modify the value at run time when needed, thus avoiding recompiling the entire system only to modify the timeout.

PR #307, opened 2024-03-18
Add parameter for timeout of plan solver

roveri-marco:
Added the possibility to set at run time the timeout for waiting for a solution from the underneath called planner.
This will fix issue #243


fmrico:
Thanks @roveri-marco . Pending of fixing CI

fmrico:
CI seems to be fixed, but it is applying old pipeline. Can you rebase/merge?

plansys2_core/include/plansys2_core/PlanSolverBase.hpp
```diff
@@ -52,7 +52,8 @@ class PlanSolverBase
   */
   virtual std::optional<plansys2_msgs::msg::Plan> getPlan(
     const std::string & domain, const std::string & problem,
-    const std::string & node_namespace = "") = 0;
+    const std::string & node_namespace = "",
+    const int solver_timeout = 15) = 0;
```
fmrico: Why `int` for the timeout? I would prefer a rclcpp::Duration? In the parameter is ok. You can document that this needs to be seconds, but internally I prefer this type.

```
using namespace std::chrono_literals;
...
const rclcpp::Duration  solver_timeout = 15s) = 0;
```

plansys2_bringup/params/plansys2_params.yaml
```diff
@@ -1,5 +1,9 @@
+planner_client:
+  ros__parameters:
+    plan_solver_timeout: 15  
 planner:
   ros__parameters:
+    plan_solver_timeout: 15
```
fmrico: Here it is okay, but I prefer double instead of int.

plansys2_planner/src/plansys2_planner/PlannerClient.cpp
```diff
@@ -28,6 +28,11 @@ PlannerClient::PlannerClient()
   node_ = rclcpp::Node::make_shared("planner_client");
 
   get_plan_client_ = node_->create_client<plansys2_msgs::srv::GetPlan>("planner/get_plan");
+
+  node_->declare_parameter("plan_solver_timeout", solver_timeout_);
```
fmrico: Here we should need to convert from double/float to rclcpp::Duration

fmrico (changes requested):

fmrico:
Everything LGTM in this PR, except the type concerns in the review

Thanks @roveri-marco

roveri-marco:
> Everything LGTM in this PR, except the type concerns in the review
> 
> Thanks @roveri-marco

I'll fix in the forthcoming days.

fmrico (approved):

fmrico:
Merging!! :rocket:

PR #316, opened 2024-08-21
[lifecycle_manager] Activate planner first

Rezenders:
I believe the planner should be activated first. 
When the domain_expert  `validate_using_planner_node` parameter is set to `true`, it waits for the planner's service `/planner/validate_domain`.


fmrico:
LGTM

Merging :rocket: 

Thanks ;)

Issue #317, opened 2024-08-22
Adding support for new planner: problems parsing pddl files


Rezenders:
Hello,

I am working on adding support for the [Symk](https://github.com/speckdavid/symk) planner to PlanSys2, mainly because I need a planner that supports derived predicates.
I created a ros package and a plansys plugin for the planner: https://github.com/Rezenders/symk_ros
And I am trying to apply it to [SUAVE](https://github.com/kas-lab/suave) exemplar with the following setup: https://github.com/Rezenders/suave_planning

However, I am having some problems with the parsing of my PDDL files. The derived predicates in the domain file don't get parsed, and some predicates in the problem file are also not getting parsed. I can check this by looking into the domain and problem files created in the `/tmp` folder.

For example, the [suave_domain_created.pddl](https://github.com/Rezenders/suave_planning/blob/main/pddl/suave_domain_created.pddl) domain file is outputted as:

```PDDL
(define (domain suave)
(:requirements :adl
 :derived-predicates :negative-preconditions
 :strips
 :typing
 )

(:types
        pipeline
        robot
        action
)

(:constants
 a_search_pipeline a_inspect_pipeline - action
    internal_error_string in_error_fr_string in_error_component_string in_error_nfr_string false_string false_boolean water_visibility recovered_string true_boolean
)

(:predicates
		(c_status ?x ?y)
		(component ?x)
		(fd_efficacy ?x ?y)
		(fd_error_log ?x ?y)
		(fd_realisability ?x ?y)
		(fg_status ?x ?y)
		(function ?x)
		(functiondesign ?x)
		(functiongrounding ?x)
		(hasnfr ?x ?y)
		(hasqaestimation ?x ?y)
		(hasqavalue ?x ?y)
		(hasvalue ?x ?y)
		(inferred-c_status ?x ?y)
		(inferred-component ?x)
		(inferred-fd_efficacy ?x ?y)
		(inferred-fd_error_log ?x ?y)
		(inferred-fd_realisability ?x ?y)
		(inferred-fg_status ?x ?y)
		(inferred-function ?x)
		(inferred-functiondesign ?x)
		(inferred-functiongrounding ?x)
		(inferred-hasnfr ?x ?y)
		(inferred-hasqaestimation ?x ?y)
		(inferred-hasqavalue ?x ?y)
		(inferred-hasvalue ?x ?y)
		(inferred-inconsistent)
		(inferred-isqatype ?x ?y)
[...]

PR #318, opened 2024-08-27
Add support for derived predicates

Rezenders:
Add support for derived predicates

Related to #317


fmrico:
Hi @Rezenders 

Very interesting feature!! But I think it is not only being able to read a pddl file, but have it full functionality that includes:

- [X] Read structure
- [ ] Show it in the PlanSys2 Terminal
- [ ] Add a `getDeriveds()` method to `DomainExpertInterface.hpp` and implement it at `DomainExpertNode.hpp` and `DomainExpertClient.hpp`
- [ ] AFAIK, a derived predicate adds a predicate to the problem when the expression is true, right? So, we should make a mechanism that, during the plan execution, for each new predicate, evaluate if a derived predicate can also be added.

Do you think we could use this PR to try to complete this list?

Thanks!!
Francisco

Rezenders:
Hello @fmrico,

Thanks for the input!

> Do you think we could use this PR to try to complete this list?

Sure :)

> Add a getDeriveds() method to DomainExpertInterface.hpp and implement it at DomainExpertNode.hpp and DomainExpertClient.hpp

What do you think should be the return type of a `getDerived` method? Should we add a new `Derived.msg`? Something like:

```
plansys2_msgs/Node inferred_predicate
plansys2_msgs/Tree preconditions
```
> AFAIK, a derived predicate adds a predicate to the problem when the expression is true, right? 

Yeah, I believe that is probably how the planner handles derived predicates. @tobiaswjohn do you have any input on this?

> So, we should make a mechanism that, during the plan execution, for each new predicate, evaluate if a derived predicate can also be added.

I am not sure I understood this part

Rezenders:
> Add a getDeriveds() method to DomainExpertInterface.hpp and implement it at DomainExpertNode.hpp and DomainExpertClient.hpp

This is supposed to be done. I still need to figure out a way to test the services.

PS: I had problems when derived predicates had `exists` conditions. It says it is not supported in PlanSys. I would also be interested in adding support for this, but I guess this can go in a separate issue/PR

https://github.com/PlanSys2/ros2_planning_system/blob/a7971965431fc5c837074e6a15059650f75645b8/plansys2_pddl_parser/src/plansys2_pddl_parser/Exists.cpp#L25-L27

fmrico:
Hi @Rezenders 

> > So, we should make a mechanism that, during the plan execution, for each new predicate, evaluate if a derived predicate can also be added.
>
> I am not sure I understood this part

I suppose that, during the execution of a plan, if the expression of a derived predicate is true, we should add the predicate (the derived predicate), right? In that case, we should think on a runtime mechanism to do it.

> PS: I had problems when derived predicates had exists conditions. It says it is not supported in PlanSys. I would also be interested in adding support for this, but I guess this can go in a separate issue/PR

Yes, it can go in a separate predicate. 

Thanks!!! :)

Rezenders:
> I suppose that, during the execution of a plan, if the expression of a derived predicate is true, we should add the predicate (the derived predicate), right? In that case, we should think on a runtime mechanism to do it.

What if we just check if the derived predicates preconditions are satisfied when `existPredicate` is called? 
Something like:

https://github.com/Rezenders/ros2_planning_system/blob/3156cba9a8a87fd3cd82bdcfb99ee6a48b0e4b31/plansys2_problem_expert/src/plansys2_problem_expert/ProblemExpert.cpp#L451-L463

I still need to test if this actually works.

Rezenders:
> I still need to test if this actually works.

I can't test with the setup I have because the planner I am using that supports derived predicates doesn't support durative actions. Related to issue #319

Rezenders:
I don't think it is ready to merge.
The evaluation is not working when `use_state` is true.

Rezenders:
> The evaluation is not working when use_state is true.

I think this can be solved by returning the derived predicates with their preconditions fulfilled in `getPredicates`. Then when the evaluation is called with `use_state` true, if the `predicates` vector was "created" by calling `getPredicates`, it should work.

fmrico:
Really good work, @Rezenders , congratulations.

Merging!!! :rocket:

PR #320, opened 2024-09-04
Support for existential preconditions

Rezenders:
Add support for existential preconditions.
I still need to adjust the evaluate function:
https://github.com/PlanSys2/ros2_planning_system/blob/a7971965431fc5c837074e6a15059650f75645b8/plansys2_problem_expert/src/plansys2_problem_expert/Utils.cpp#L29

I am thinking about solving this by getting all instances with the problem client and evaluating the exists tree for all possible instance combinations until one is satisfied. 

related to #317

### DIFF
--- a/plansys2_pddl_parser/src/plansys2_pddl_parser/Ground.cpp
+++ b/plansys2_pddl_parser/src/plansys2_pddl_parser/Ground.cpp
@@ -38,7 +38,7 @@ plansys2_msgs::msg::Node::SharedPtr Ground::getTree( plansys2_msgs::msg::Tree & 
     node->name = name;
     for ( unsigned i = 0; i < params.size(); ++i ) {
         plansys2_msgs::msg::Param param;
-        if (i < replace.size()) {
+        if (i < replace.size() && params[i]<replace.size()) {
           if (params[i] >= 0) {
             // param has a variable value; replace by action-args
             param.name = replace[params[i]];
--- a/plansys2_msgs/msg/Node.msg
+++ b/plansys2_msgs/msg/Node.msg
@@ -11,24 +11,25 @@ uint8 FUNCTION = 6
 uint8 EXPRESSION = 7
 uint8 FUNCTION_MODIFIER = 8
 uint8 NUMBER = 9
+uint8 EXISTS = 10
 
 # Expression types
-uint8 COMP_EQ = 10
-uint8 COMP_GE = 11
-uint8 COMP_GT = 12
-uint8 COMP_LE = 13
-uint8 COMP_LT = 14
-uint8 ARITH_MULT = 15
-uint8 ARITH_DIV = 16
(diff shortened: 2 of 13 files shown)


Rezenders:
I included the evaluation for the `exists` clause. It checks if any combination of instances fulfills the `exists` condition. It is not optimal, but it is the only solution I could think of. It only works when `use_state` is `false`

https://github.com/PlanSys2/ros2_planning_system/blob/bf54092e9eccdaf34aae1237bd7c56536a0b1809/plansys2_problem_expert/src/plansys2_problem_expert/Utils.cpp#L305-L344

Maybe we can check the type of the parameters to reduce the number of instance combinations

plansys2_pddl_parser/src/plansys2_pddl_parser/Ground.cpp
```diff
@@ -38,7 +38,7 @@ plansys2_msgs::msg::Node::SharedPtr Ground::getTree( plansys2_msgs::msg::Tree &
     node->name = name;
     for ( unsigned i = 0; i < params.size(); ++i ) {
         plansys2_msgs::msg::Param param;
-        if (i < replace.size()) {
+        if (i < replace.size() && params[i]<replace.size()) {
```
fmrico, line 41: I think linters are not working here, because you need spaces around the operator `<`

fmrico (changes requested):

plansys2_pddl_parser/src/plansys2_pddl_parser/Ground.cpp
Rezenders: Yeah, linters are deactivated: https://github.com/PlanSys2/ros2_planning_system/blob/ce2c82832e448ed0dde771d2ef50bcb206e88f82/plansys2_pddl_parser/CMakeLists.txt#L62-L63

I turned the linters on and fixed the lintering issues and got cpplint to pass, but I can't make both uncrustify and cpplint pass at the same time, there are some conflicts. I have no clue why this only happens in this package. Could there be a difference because the header files are `.h` instead of `.hpp`?

Rezenders: Yeah, this problem was because the header files have the `.h` extension, I changed them to `.hpp` and fixed all lintering issues and activated the lintering test. I am not sure if it was the best approach to include these changes in this PR, let me know if you prefer me to create a separate PR.

Rezenders: There were conflicts. I will open a new PR with the linter changes.

fmrico (approved):

fmrico:
Hi @Rezenders 

I think everything is ok. Thanks!!

Merging!!! 🚀

PR #338, opened 2024-12-17
[WIP] Replanning

fmrico:
Dear all,

This WIP PR provides PlanSys2 with replanning capability and some refactoring.

- [x] Basic replanning: If a controller sends a new plan, the previous plan is canceled, and the new plan is started.
- [x] Smooth replanning: If a new plan is sent, the actions that have already started are continued if they are the first to be executed in the new plan.
- [ ] Improved cancelation: Actions to be canceled have a mechanism to safely terminate, even delaying the start of the new plan.

I would like everyone to go over the ongoing work here.

Best


fmrico:
I am going to merge this because the last point requires more time to think in a good solution...

PR #341, opened 2025-01-14
[WIP] Multiple plans and replanning stability

fmrico:
Dear all,

This is a big PR that contains:

1. Improvements in the stability of PlanSys when switching from one plan to another. Some stability problems were introduced in #338, but they only appear when using the new replanning functionality.
2. There is a new service `planner/get_plan_array[plansys2_msgs::srv::GetPlanArray]`, which returns all the plans calculated in parallel by the different plugins. For example, if you have this params file:

```
planner_client:
  ros__parameters:
    plan_solver_timeout: 1.0  
planner:
  ros__parameters:
    plan_solver_timeout: 1.0  
    plan_solver_plugins: ["POPF1", "POPF2", "POPF3", "POPF4", "OPTICS"]
    POPF1:
      plugin: "plansys2/POPFPlanSolver"
      arguments: "-h -E -A"
      output_dir: "/tmp/POPF1"
    POPF2:
      plugin: "plansys2/POPFPlanSolver"
      arguments: "-h -E"
      output_dir: "/tmp/POPF2"
    POPF3:
      plugin: "plansys2/POPFPlanSolver"
      arguments: "-h"
      output_dir: "/tmp/POPF3"
    POPF4:
      plugin: "plansys2/POPFPlanSolver"
      arguments: ""
      output_dir: "/tmp/POPF4"
    OPTICS:
      plugin: "plansys2/OPTICPlanSolver"
      arguments: "-N"
      output_dir: "/tmp/OPTICS"
executor:
  ros__parameters:
    bt_builder_plugin: "SimpleBTBuilder"
```

, this service returns all the plans calculated by all the `plan_solver_plugins` entries calculated within the timeout.

3. Related to the last point, now `planner/get_plan[plansys2_msgs::srv::GetPlan]` return the sorter of the plans calculated by all the plugins.

I am finishing testing this, so it will probably not be merged immediately. Meanwhile, I would like to know what you think about these changes.

Best


fmrico:
BTW, I have updated the OPTICS planner and PlanSys2 plugin here https://github.com/sscpac/optic_planner/pull/14

gregtkogut:
This is great!  We were recently talking about how great it would be to run multiple PDDL planners in parallel and then have some sort of logic on top to pick which plan to run - from something as simple as minimum temporal span to conceivably any combination of plan characteristics.    

One nice-to-have would be planner timeout parameters per planner instance, as the runtime profile of different planners can vary greatly.   And another future capability might be to return plans as they come in vs. returning them all at once.  That could let quick planners return their results right away while long-running planners can keep churning.  That way your quickest planner isn't held back by your slowest planner.

Nit:  It's `OPTIC` vs. `OPTICS`.

roveri-marco:
I've been working on online planning, where the robotic agents start executing a (partial) plan without having completed the generation of the full plan. It was integrated in an old version of plansys2 and it would be good to integrate it in the latest version. For this purpose we used a modified temporal planner (unfortunately written in Java) and defined a protocol to have the planner and plansys2 execution to communicate and to decide whether to continue with the plan in execution, abandon it, and to replan. The whole project is available here: https://github.com/RTI-BDI/ros2bdi_build_workspace

fmrico:
@gregtkogut 

> One nice-to-have would be planner timeout parameters per planner instance, as the runtime profile of different planners can vary greatly. 

It is not hard to implement, but we should find a use case. I don't see it so useful. This service needs a global timeout.

> And another future capability might be to return plans as they come in vs. returning them all at once. That could let quick planners return their results right away while long-running planners can keep churning. That way your quickest planner isn't held back by your slowest planner.

This is an interesting idea, but we must consider how to implement it. The Planner Expert could have an action server that starts continuous plan calculation processes and sends the plans as action feedback to the client as they are ready.

An interesting question is to see if a plan is useful if it takes a long time to calculate since if it takes a long time when it is ready, the problem status will be different.

@roveri-marco 

> I've been working on online planning, where the robotic agents start executing a (partial) plan without having completed the generation of the full plan. It was integrated in an old version of plansys2 and it would be good to integrate it in the latest version. For this purpose we used a modified temporal planner (unfortunately written in Java) and defined a protocol to have the planner and plansys2 execution to communicate and to decide whether to continue with the plan in execution, abandon it, and to replan. The whole project is available here: https://github.com/RTI-BDI/ros2bdi_build_workspace

Can't this be compared to calculating complete plans but constantly replanning, in such a way that only the first actions of each plan are executed?

In any case, the intelligence of these applications resides in the controller, and the idea is that PlanSys2 offers you the necessary functionality to be able to carry it out. Don't you think?

roveri-marco:
> Can't this be compared to calculating complete plans but constantly replanning, in such a way that only the first actions of each plan are executed?

This sub-optimal approach works only if the plan is sequential. If there are overlapping temporal actions (supported by the BT construction) and possible depending on how the PDDL domain and problems have been formulated it does not work, since one has to deal with possible intermediate states where the started actions have not been terminated, and termination depends on some other actions still to be executed.

> In any case, the intelligence of these applications resides in the controller, and the idea is that PlanSys2 offers you the necessary functionality to be able to carry it out. Don't you think?

Yes, but some changes need to be done (indeed we did some minor internal modifications).

gregtkogut:
@fmrico, yeah, on further thought I agree a single overarching timeout makes the most sense right now.

fmrico:
Hi,

More work in this PR to improve the stability of the plan executions when using replanning:
- [x] Executor refactored
- [x] Improve Planner Plugins to cancel the process of planning
- [x] A new method in ProblemClient (`getProblemWithTimestamp`) to get the problem with the timestamp for checking validity 
- [x] Option by parameter to execute PlanSys2 (only monolithic) in Real Time
- [x] A new BT plugin for the actions ` RestoreAtStartEffect`

PR #342, opened 2025-01-21
Fixed problem for debugging in rqt

Juancams:
HI!!!

There was a problem with the rqt display, and it was that the plan was visible but not the execution of it (nor the colors).

The reason was that the plan status was published first and then the plan, and in the rqt node all the status was deleted when a new plan arrived.

There were two solutions:
- Publish the plan first and then the status.
- Publish the plan at the beginning or when it changes.

I have chosen this second action because it would be published less, and also because when viewing it the rqt window was updating much more, and it looked ugly :(

Greetings ^^


fmrico:
Hi @Juancams 

As we discussed, there are some drawbacks in your fix. I have fixed, maybe not totally, this in #341, by changing the way the plan and updates are managed in RQTPlugin.

I am closing this. Thanks anyway for sharing your vision.

PR #353, opened 2025-03-26
Closes #352

### LINKED ISSUE #352: [POPF Plugin] Error when getting a (successful) plan
Using POPF, when the goal is already satisfied in the current set of predicates, the planner will output "Solution Found", but the list of actions in the plan will be empty (for obvious reasons).

With the [current implementation](https://github.com/PlanSys2/ros2_planning_system/blob/3fc9e946067c75169772851c5d762d323efd5383/plansys2_popf_plan_solver/src/plansys2_popf_plan_solver/popf_plan_solver.cpp#L159-L163), the `getPlan` function returns an empty optional, which results in an error from the Planner module stating that the plan could not be found.

Is there a reason to have this implementation? Should the `getPlan` put a value in the optional with an empty plan?

I understand that one should not set a goal that is already satisfied, but it may happen and the error is rather confusing. I believe returning an empty plan transmits the message that "there is nothing to be done", without returning an error. If the error is needed somewhere else, then I would like to understand the reasons.

aquintan4:
Fix POPF plugin issue when retrieving a successful plan (#352) and add a message in the terminal for plans with no actions (but not empty).

### DIFF
--- a/plansys2_popf_plan_solver/src/plansys2_popf_plan_solver/popf_plan_solver.cpp
+++ b/plansys2_popf_plan_solver/src/plansys2_popf_plan_solver/popf_plan_solver.cpp
@@ -156,7 +156,7 @@ POPFPlanSolver::parse_plan_result(const std::string & plan_path)
     plan_file.close();
   }
 
-  if (solution && !plan.items.empty()) {
+  if (solution) {
     return plan;
   } else {
     return {};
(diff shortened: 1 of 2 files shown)


fmrico:
LGTM!! Merging :rocket: 

Thanks @aquintan4

PR #354, opened 2025-03-28
Fix string predicate parsing to handle cases with no parameters

eshan-savla:
This pull request includes a minor change to the `plansys2_pddl_parser` module. The change involves the `fromStringPredicate` function in the `Utils.cpp` file, adding checks to handle edge cases where the predicates have no parameters.

* [`plansys2_pddl_parser/src/plansys2_pddl_parser/Utils.cpp`](diffhunk://#diff-4673b16b2bf073681d59660e7d2d206f80c65307c5bad20774d5bb5f7a6305baR944-R951): Added a check to remove a closing parenthesis from the end of the first token and an early return if the token list size is one, indicating no parameters are provided.


eshan-savla:
Would this also be considered a bug fix? If so, I can push the respective change from the humble branch I am working on

roveri-marco:
Can you also add a test case that was failing before the patch and that pass after the patch?

fmrico:
> Can you also add a test case that was failing before the patch and that pass after the patch?

Yes, please, @eshan-savla . A test will be very helpful to merge this PR

eshan-savla:
Hello all, I have added the requested test cases and rebased the fix on those commits. Now you should be able to also manually test whether these tests fail before my fix.

fmrico:
Thanks @eshan-savla 

Merging!!! :rocket:

PR #369, opened 2025-06-13
Improvements on BTAction

ajtudela:
I made some improvements to the BTActions:
- I added ```server_timeout``` as a parameter.
- I split timeouts between server and wait, and I added a new ```wait_for_service_timeout``` parameter.
- I removed the rate from the ActionExecutorClient constructor to simplify its usage, it was only used in bt_actions, and is now a fully set parameter. I renamed the internal member to 'period_' for clarification. The default is 200ms / 5.0Hz.
- I added a new utils to function (from Davide in Nav2) to handle with the setting-getting of the ports / blackboard.
- The new parameters are set in the blackboard of the BT. Adding to the blackboard it makes all Nav2 nodes compatible with PlanSys2.

### DIFF
--- a/plansys2_bt_actions/include/plansys2_bt_actions/BTAction.hpp
+++ b/plansys2_bt_actions/include/plansys2_bt_actions/BTAction.hpp
@@ -50,6 +50,15 @@ public:
   explicit BTAction(const std::string & action);
 
   /**
+   * @brief Constructor for the BTAction.
+   *
+   * @param[in] action Name of the action this executor handles.
+   * @param[in] rate Execution rate for the action.
+   */
+  [[deprecated("Use BTAction(const std::string & action) instead")]]
+  BTAction(const std::string & action, const std::chrono::nanoseconds & rate);
+
+  /**
    * @brief Get the name of the action.
    *
    * @return std::string The action name.
--- a/plansys2_bt_actions/include/plansys2_bt_actions/BTAction.hpp
+++ b/plansys2_bt_actions/include/plansys2_bt_actions/BTAction.hpp
@@ -152,6 +152,9 @@ private:
   // The timeout value for waiting for a service to response
   std::chrono::milliseconds wait_for_service_timeout_;
 
+  // Duration for each iteration of BT execution
+  std::chrono::milliseconds bt_loop_duration_;
+
   // A regular, non-spinning ROS node that we can use for calls to the action client
   rclcpp::Node::SharedPtr client_node_;
 };
--- a/plansys2_bt_actions/include/plansys2_bt_actions/BTAction.hpp
+++ b/plansys2_bt_actions/include/plansys2_bt_actions/BTAction.hpp
@@ -146,6 +146,12 @@ private:
   std::unique_ptr<BT::MinitraceLogger> bt_minitrace_logger_;
   // Groot2 monitor
   std::unique_ptr<BT::Groot2Publisher> groot_monitor_;
+
+  // Default timeout value while waiting for response from a server
+  std::chrono::milliseconds default_server_timeout_;
+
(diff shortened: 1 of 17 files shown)


plansys2_bt_actions/src/plansys2_bt_actions/BTAction.cpp
```diff
@@ -54,21 +54,42 @@ BTAction::on_configure(const rclcpp_lifecycle::State & previous_state)
...
+  // Create a regular, non-spinning ROS node that we can use for calls to the action client
+  client_node_ = std::make_shared<rclcpp::Node>("_");
+
+  // Create the blackboard that will be shared by all of the nodes in the tree
   blackboard_ = BT::Blackboard::create();
-  blackboard_->set("node", shared_from_this());
+
+  // Put items in the blackboard
+  blackboard_->set<rclcpp::Node::SharedPtr>("node", client_node_);  // NOLINT
+  blackboard_->set<std::chrono::milliseconds>("server_timeout", default_server_timeout_);  // NOLINT
```
fmrico: Why `// NOLINT`? I prefer to fix sthe tyle

plansys2_bt_actions/include/plansys2_bt_actions/BTAction.hpp
```diff
@@ -34,9 +34,7 @@ namespace plansys2
 class BTAction : public plansys2::ActionExecutorClient
 {
 public:
-  explicit BTAction(
```
fmrico: I prefer to maintain both constructors for backward compatibility. You can add a warning of deprecation in the old constructor.

plansys2_bt_actions/test/behavior_tree/Move.cpp
```diff
@@ -32,7 +32,7 @@ Move::Move(
   const BT::NodeConfig & conf)
 : plansys2::BtActionNode<test_msgs::action::Fibonacci>(xml_tag_name, action_name, conf)
 {
-  rclcpp_lifecycle::LifecycleNode::SharedPtr node;
+  rclcpp::Node::SharedPtr node;
```
fmrico: This is very sensible. Currently, these BT nodes are used in action performers, which are cascade_lifecycle...

plansys2_executor/test/unit/executor_test.cpp
```diff
@@ -224,13 +224,16 @@ TEST(executor, action_executor_client)
   auto test_node = rclcpp_lifecycle::LifecycleNode::make_shared("test_node");
   auto aux_node = rclcpp_lifecycle::LifecycleNode::make_shared("aux_node");
```
fmrico, line 249: As I said in the previous comment, let's maintain both constructors, so duplicate the test and modify the new test, and maintain this.

plansys2_bt_actions/test/behavior_tree/Move.cpp
ajtudela: I understand the need for the main nodes to be lifecycle nodes. However, the node within the BTActionNode does not have to be a lifecycle node because the action is executed when the ActionExecutorClient instructs it to. In fact, the advantages of lifecycle nodes (configure, activate, etc.) are never utilized in the BTActionNode.

A node client inside the BTActionNode works the same way: 
https://github.com/PlanSys2/ros2_planning_system/compare/rolling...grupo-avispa:ros2_planning_system:improve_bt_actions#diff-d7c4293a0e6ee3223f93bd6fcd115bb5fe9b52f2049ac5e266010e9a7c47fba8R78-R79

ajtudela:
I added a ```deprecated```attribute to the old constructor and a new test that use it. Other tests are updated with the new constructor.
I reverted the rclcppp_lifecycle inside the BT.

fmrico:
LGTM

Merging!!! :rocket:

PR #371, opened 2025-07-03
add plansys2::SequentialBTBuilder plugin

Rezenders:
Add plansys2::SequentialBTBuilder plugin.

This plugin creates a Behavior Tree to execute the actions of a plan generated by the planner in the exact order they appear. In contrast to the SimpleBTBulder that attempts to parallelize the plan actions.

For example, the plan:

```
(action1 param1)
(action2 param2)
(action3 param3)
```

Would be converted to the BT:

```xml
<root BTCPP_format="4" main_tree_to_execute="MainTree">
  <BehaviorTree ID="MainTree">
    <Sequence name="MainSequence">
      <Sequence name="(action1 param1):0">
        <ApplyAtStartEffect action="(action1 param1):0"/>
        <ReactiveSequence name="(action1 param1):0">
            <CheckOverAllReq action="(action1 param1):0"/>
            <ExecuteAction action="(action1 param1):0"/>
        </ReactiveSequence>
        <CheckAtEndReq action="(action1 param1):0"/>
        <ApplyAtEndEffect action="(action1 param1):0"/>
      </Sequence>
      <Sequence name="(action2 param2):0">
        <WaitAction action="(start_robot bluerov):0"/>
        <ApplyAtStartEffect action="(action2 param2):0"/>
        <ReactiveSequence name="(action2 param2):0">
            <CheckOverAllReq action="(action2 param2):0"/>
            <ExecuteAction action="(action2 param2):0"/>
        </ReactiveSequence>
        <CheckAtEndReq action="(action2 param2):0"/>
        <ApplyAtEndEffect action="(action2 param2):0"/>
      </Sequence>
      <Sequence name="(action3 param3):0">
        <WaitAction action="(start_robot bluerov):0"/>
        <WaitAction action="(action2 param2):0"/>
        <ApplyAtStartEffect action="(action3 param3):0"/>
        <ReactiveSequence name="(action3 param3):0">
            <CheckOverAllReq action="(action3 param3):0"/>
            <ExecuteAction action="(action3 param3):0"/>
        </ReactiveSequence>
        <CheckAtEndReq action="(action3 param3):0"/>
        <ApplyAtEndEffect action="(action3 param3):0"/>
      </Sequence>
    </Sequence>
  </BehaviorTree>
</root>
```

This BT builder plugin is useful when:
- The assumptions of the other BT builder plugins do not hold. For example, the SimpleBTBuilder plugin cannot be used when there are existential predicates in the action's preconditions
- You want to guarantee that the plan will be executed exactly as it was generated. This can be a requirement in some applications. @alex-quin-gabriel is working on an application that has this requirement. @alex-quin-gabriel, if you have time, could you write a brief description of the use case, please?


fmrico:
Great contribution!!! 

I will accept this as it is, but first I would be very thankful if you could edit the first message in this PR to explain the benefits of this BTBuilder, how it works, and in what cases any user could want to use it. Not very long, but enough to be useful.

This could go directly to the documentation.

Thanks for your work!!

Rezenders:
> I will accept this as it is, but first I would be very thankful if you could edit the first message in this PR to explain the benefits of this BTBuilder, how it works, and in what cases any user could want to use it. Not very long, but enough to be useful.

Done. 

When this PR gets merged, I can update the plugins page in the docs.

fmrico:
Merged!!!

🚀

alex-quin-gabriel:
I want to use a reasoner to take care of the interdependencies between individual steps, let it come up with a sequence of actions (of processing data) and then have that realized in the system....without some other part changing that sequence.

roveri-marco:
But, are you considering timed actions or untimed ones? Sequencing can be forced in the PDDL model, but it depends on the generated plan. What happens if the plan can be parallelized since actions can be executed in parallel?
This has to be properly handled... It is not just a matter of having the sequential BT builder. In my opinion it depends on the PDDL model.

I'm thinking something similar, but then I will switch to a pure untimed PDDL model, where actions are executed sequentially (although they can be parallelized, but then i is not clear the duration and dependencies).

PR #330, opened 2024-10-24
Improvements to the evaluate function focusing on optimizing derived predicates and existential preconditions resolution

Rezenders:
This PR is a draft to solve issue #328. 

I implemented the algorithm described by @tobiaswjohn for most of the Plansys2 Node types.

This version is much faster than the previous version in the main branch. It takes around 37643ms to solve all derived predicates in [this PDDL formulation](https://github.com/Rezenders/ros2_planning_system/blob/new_evaluation/plansys2_problem_expert/test/pddl/domain_derived.pddl), whereas it wouldn't finish with the previous version of the algorithm.

The evaluate function could be further improved by:
-  by feeding the already assigned variables to the eval function
- with an algorithm that checks the type of the derived predicate parameters and finds out the possible types for the variables in the derived predicate preconditions, which could reduce the search space significantly in certain cases

I still need to adjust the other packages to make it compatible with the new evaluation and problem expert.

I didn't adjust the evaluation for anything related to `functions`. If you think this approach would be mergeable when complete, I can adjust it to handle `functions`. But right now, I don't need it for my project, so I will focus on finishing what I need first.

Issue #328, opened 2024-10-04
Inefficient existential preconditions resolution


Rezenders:
Hi @fmrico,

Unfortunately, the solution I proposed for existential predicates is inefficient. It is literally exploring the whole state space to check if the predicates are satisfied. This can easily explode. For example, if there are 75 instances and an existential precondition has 5 variables, this results in 75^5 possible combinations to check, which is unfeasible. 

A concrete example (complete formulation can be found [here](https://github.com/Rezenders/suave_planning/tree/main/pddl)): 
```pddl
	(:derived (inferred-Fg_status ?fg ?IN_ERROR_NFR_string)
		(and
			(= ?IN_ERROR_NFR_string IN_ERROR_NFR_string)
			(exists (?mqa ?eqa ?mqav ?fd ?eqav)
 				(and
					(inferred-FunctionGrounding ?fg)
					(inferred-HasQAvalue ?fg ?mqa)
					(inferred-IsQAtype ?eqa water_visibility)
					(inferred-HasValue ?mqa ?mqav)
					(inferred-TypeFD ?fg ?fd)
					(inferred-HasValue ?eqa ?eqav)
					(inferred-HasQAestimation ?fd ?eqa)
					(inferred-IsQAtype ?mqa water_visibility)
					(inferred-FunctionDesign ?fd)
					(inferred-QAvalue ?mqa)
					(inferred-QAvalue ?eqa)
					(inferred-LessThan ?mqav ?eqav)
				)
			)
 		)
 	)
```

We need to come up with a mechanism to prune the possible variable combinations. I don't have any good ideas so far.

tobiaswjohn:
The general procedure should be to start from the state, i.e. known facts and build the set of all derived facts from that. This applies to the existential quantification and the derived predicates.

I assume, that there is a set `S` that contains all atomic formulas of the form `(p a_1 ... a_n)` that are true in the current state

- notations:
    - variables are named `?x_1` until `?x_n`
    - object names are `a_1` until `a_n` 
    - derived predicates have the form `(:derived (p ?x_1 ... ?x_n) P')`, where the predicate `(p ?x_1 ... ?x_n)` is derived, if the condition `P'` is satisfied (where `?x_1,...,?x_n` occur freely in `P'`)

## Algorithm for derived predicates
[...]

PR #372, opened 2025-07-03
Plansys2 core

Rezenders:
Related to PR #330. I am breaking up that PR into smaller PRs.

Add `==` operator and `hash` functions to classes:
- Instance
- Predicate
- Function

Add: 
- `Derived` class to represent a `plansys2_msgs::msg::Derived`
- `Action` class to represent `plansys2_msgs::msg::Action`
- `DurativeAction` class to represent `plansys2_msgs::msg::DurativeAction`
- `ActionVariant` class to hold either an `Action` or `DurativeAction` (moved from ActionExecutor.hpp and extended)
- `NodeVariant` class to hold either a `Predicate`, `Function`, `Derived`, or `ActionVariant`. Instances of this class are the nodes of the derived graph
- `DerivedResolutionGraph` class to represent the derived predicates dependencies as described in #328
- `State` class to represent the planning state. That is, `<Instances, Functions, Predicates, InferredPredicates, DerivedResolutionGraph>`
- `State` message

Modified:
- `ComputeBT`, `ExecutorNode` and `simple_bt_builder` to match the new `ActionVariant` class


Rezenders:
@fmrico this PR introduces new classes and modifications that were needed to implement the resolution of derived predicates (issue #328), but it doesn't add any new functionalities.
I'm not sure whether it makes sense to create a PR that adds new classes that are still not used. I want to split PR #330 into smaller and "more reviewable" PRs. Let me know if I should change this PR to include the changes that use these new classes.

fmrico:
Hi @Rezenders 

I am carefully checking this PR, because this is the very core of PlanSys2. @roveri-marco , if you can double-check this, I would be very grateful.

As far as I have reviewed, LGTM, except one concern with your current PR: if the changes you are making in `plansys2_core` are only going to be used in `plansys2_executor`, I would prefer to keep everything within `plansys2_executor`. You can create a `plansys2_executor_types` or `plansys2_executor_core` directory under both `src` and `include` to place them there.

Of course, it is not easy to understand the final goal of this PR, since, as it stands, you are just introducing new types for use in a future PR. I think you should include more commits here, because all these changes should be integrated atomically. Otherwise, many of them would not make sense on their own. At least the minimal parts to consider that this PR makes a contribution. At the very least, enough to consider this PR a worthwhile contribution that deserves to be merged.

Thanks!!

roveri-marco:
@fmrico I'll have a look in these days.

plansys2_core/include/plansys2_core/Action.hpp
```diff
@@ -0,0 +1,213 @@
...
+
+    return seed;
+  }
+};
+
+template<>
+struct hash<plansys2::DurativeAction>
+{
+  std::size_t operator()(const plansys2::DurativeAction & action) const noexcept
+  {
```
roveri-marco, line 96: Durative actions have a duration (that in PDDL might not be constant). How do you deal with this?

plansys2_core/include/plansys2_core/Action.hpp
```diff
@@ -0,0 +1,213 @@
...
+    return std::visit(
+      [](auto && arg) {return std::hash<std::decay_t<decltype(arg)>>{}(arg);}, *action_);
+  }
+
+  std::string get_action_string() const;
+
+  std::string get_action_name() const;
+
+  std::vector<plansys2_msgs::msg::Param> get_action_params() const;
```
roveri-marco, line 182: Again, what about the duration of a durative action? How do you deal with this?

plansys2_core/include/plansys2_core/DerivedResolutionGraph.hpp
```diff
@@ -0,0 +1,226 @@
...
+#include "plansys2_core/NodeVariant.hpp"
+#include "plansys2_core/Types.hpp"
+#include "plansys2_msgs/msg/derived.hpp"
+#include "plansys2_msgs/msg/node.hpp"
+#include "plansys2_msgs/msg/tree.hpp"
+
+namespace plansys2
+{
+
+class DerivedResolutionGraph
```
roveri-marco, line 74: It would be appreciated to document what's the purpose of this class!

plansys2_core/include/plansys2_core/State.hpp
```diff
@@ -0,0 +1,355 @@
...
+#include "plansys2_core/Types.hpp"
+#include "plansys2_msgs/msg/derived.hpp"
+#include "plansys2_msgs/msg/node.hpp"
+#include "plansys2_msgs/msg/param.hpp"
+#include "plansys2_msgs/msg/state.hpp"
+
+namespace plansys2
+{
+
+class State
```
roveri-marco, line 45: I can think of this class, but it would be important to document the class more precisely.

plansys2_core/include/plansys2_core/Action.hpp
Rezenders: The DuraticeAction class inherits from[ `plansys2_msgs::msg::DurativeAction.msg`](https://github.com/PlanSys2/ros2_planning_system/blob/rolling/plansys2_msgs/msg/DurativeAction.msg), which doesn't contain any duration information. That is why it was not included. 

> that in PDDL might not be constant

By this, I assume you mean that in a given plan the duration of the actions might change due to the grounded parameters of the action, something like: https://planning.wiki/ref/pddl21/domain#inequality-value
In this case, this information is only available when the plan is generated and the planner plugin passes this information with a [PlanItem.msg](https://github.com/PlanSys2/ros2_planning_system/blob/rolling/plansys2_msgs/msg/PlanItem.msg) not with a [ `plansys2_msgs::msg::DurativeAction.msg`](https://github.com/PlanSys2/ros2_planning_system/blob/rolling/plansys2_msgs/msg/DurativeAction.msg)

plansys2_core/include/plansys2_core/Action.hpp
Rezenders: Check my previous comment: https://github.com/PlanSys2/ros2_planning_system/pull/372#discussion_r2194725113

This info is captured in the[ `ActionExecutionInfo` ](https://github.com/Rezenders/ros2_planning_system/blob/84de15396d7c7ff841179d563219138dcd739aad/plansys2_executor/include/plansys2_executor/ActionExecutor.hpp#L290-L302) class

plansys2_core/include/plansys2_core/Action.hpp
roveri-marco: Yes, an this means it is an expression to be evaluated! In my opinion, it shall be handled otherwise I do not see the reason to have all the other elements of a durative action but the duration.

roveri-marco:
I've not yet finished reviewing the code... A very busy period! Sorry!

plansys2_core/include/plansys2_core/Action.hpp
fmrico: Yes, this is an interesting discussion: I have never included duration in durative actions because I have always assumed that the action duration in the real world can't be known beforehand. Talks with @roveri-marco have opened my mind with respect to this, as maybe in some cases, if times are not met, the plan should fail. In any case, @Rezenders is making some changes assuming the current position: the time of durative actions is not critical (at the moment). @Rezenders , are there any changes in this PR that remove the possibility of taking these times into account in the future?

fmrico:
Ok. The code LGTM, knowing that it is a refactoring needed for an incoming improvement. Apart from passing the automatic tests, I manually tested it, and everything works fine. Let's focus on the next PRs, @roveri-marco . Thanks for the reviews you could do :wink: 

I am merging this PR to let @Rezenders continue his work here. 

Thanks, Gustavo!!!
:rocket:

PR #376, opened 2025-07-28
Add Predicate-Based Behavior Tree Executor

ajtudela:
This PR introduces a new predicate-based behavior tree executor (```PredicateBTExecutor```) for executing behavior trees, enabling sub-symbolic checks of predicates.

* The PredicateBTExecutor integrates with the ```WaitAtStartReq``` BT node and is responsible for checking if predicates exist using the ```ProblemClient``` and executing predicate-based behavior trees for sub-symbolic validation.
* New parameters have been added (```predicate_plugins```, ```predicates_bt_xml```) similar to BT Actions to retrieve the names and XML templates of the behavior trees.
* These XML templates are passed to the PredicateBTExecutor through the Blackboard for seamless integration.

I have successfully tested this implementation on our robots and believe it's worth discussing for potential integration.

### DIFF
--- a/plansys2_bringup/launch/plansys2_bringup_launch_monolithic.py
+++ b/plansys2_bringup/launch/plansys2_bringup_launch_monolithic.py
@@ -34,7 +34,6 @@ def generate_launch_description():
     start_action_bt_file = LaunchConfiguration('start_action_bt_file')
     end_action_bt_file = LaunchConfiguration('end_action_bt_file')
     bt_builder_plugin = LaunchConfiguration('bt_builder_plugin')
-    predicate_bt_paths = LaunchConfiguration('predicate_bt_paths')
 
     declare_model_file_cmd = DeclareLaunchArgument(
         'model_file',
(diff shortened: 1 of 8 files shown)


fmrico:
Can you explain the advantages of this approach and the problem that this PR wants to fix? Also I need to see why to use a BT for checking predicates

ajtudela:
The PR attempts to bridge the gap between symbolic and subsymbolic information by introducing a mechanism for verifying predicates using behaviour trees. Each predicate is linked to a BT node, and when the executor needs to check whether the symbolic predicates have been met, it also executes the associated BT.

roveri-marco:
That's great! I was indeed planning to do something similar. Does it also work with numeric fluents? What's thei role with the effects of an action?

ajtudela:
> That's great! I was indeed planning to do something similar. Does it also work with numeric fluents? What's thei role with the effects of an action?

It is currently just a proof of concept, with only the at-start requirements implemented. I'm no expert on planning and I'm not exactly sure how the effect works.

fmrico:
We should wait for CI in this PR

PR #385, opened 2025-10-08
Allow `rolling` to build against ROS Jazzy

gavanderhoorn:
As per title, and as discussed in #384.

This lets me follow the regular build & installation instructions and build `rolling` against a Jazzy install.

Note: I've only build-tested this. I've not actually run anything.

### DIFF
--- a/plansys2_tools/include/rqt_plansys2_knowledge/RQTKnowledge.hpp
+++ b/plansys2_tools/include/rqt_plansys2_knowledge/RQTKnowledge.hpp
@@ -29,7 +29,12 @@
 #include <map>
 #include <memory>
 
+#include "rclcpp/version.h"
+#if RCLCPP_VERSION_GTE(29, 0, 0)
 #include "rqt_gui_cpp/plugin.hpp"
+#else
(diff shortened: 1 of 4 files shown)


plansys2_tools/include/rqt_plansys2_knowledge/RQTKnowledge.hpp
```diff
@@ -29,7 +29,12 @@
 #include <map>
 #include <memory>
 
+#include "rclcpp/version.h"
+#if RCLCPP_VERSION_GTE(29, 0, 0)
```
gavanderhoorn, line 33: this checks `rclcpp`'s version, as `rqt_gui_cpp` doesn't currently install a `version.h` header.

An alternative could be to use `__has_include(..)`, which would make the conditional check for the existence of `rqt_gui_cpp/plugin.hpp` directly, instead of relying on 'circumstantial' evidence (`rclcpp` being of a specific minimum version).

fmrico:
Thanks @gavanderhoorn 

Merging!!! :rocket:

PR #389, opened 2025-11-21
Support custom messages in BTAction

avalen2022:
Use  from the blackboard for feedback and finish messages instead of generic ones.


fmrico:
Hi @avalen2022 

LGTM - merging!!! :rocket:

Thanks

fmrico:
I have just noticed that the target is `jazzy-devel`, instead of rolling. It doesn't hurt `jazzy-devel` branch, but all the developments must be done in `rolling` to be definitely incorporated in PlanSys2.

@avalen2022 , may you do again the changes in `rolling`and do the PR again, targeting that branch?

Issue #391, opened 2025-12-14
types_test derived_predicates_get_normalized fails


fmrico:
Hi,

The test `TEST(types_test, derived_predicates_get_normalized)` in `plansys2_core` now fails, even though it passed when the PR was accepted. I deactivated them to pass the CI, but we should reactivate them. @Rezenders did this part. May you take a look at the marked commented lines, please:

```
TEST(types_test, derived_predicates_get_normalized)
{
  plansys2::Derived inferredA;
  inferredA.predicate = parser::pddl::fromStringPredicate("(inferredA ?a)");
  inferredA.preconditions = parser::pddl::fromString("(and (predicateA ?a))");
  auto norm_inferredA = inferredA.getNormalizedDerived();

  // ASSERT_NE(inferredA, norm_inferredA);  // ToDo: Fix this test @Rezenders

  plansys2::Derived inferredAconstant;
  inferredAconstant.predicate = parser::pddl::fromStringPredicate("(inferredA ?a)");
  inferredAconstant.preconditions = parser::pddl::fromString("(and (predicateA a))");
  auto norm_inferredAconstant = inferredAconstant.getNormalizedDerived();

  ASSERT_EQ(norm_inferredA.predicate.parameters[0].name, "?0");
  ASSERT_EQ(norm_inferredA.preconditions.nodes[1].parameters[0].name, "?0");

  // ASSERT_NE(inferredAconstant, norm_inferredAconstant);  // ToDo: Fix this test @Rezenders

  ASSERT_EQ(norm_inferredAconstant.predicate.parameters[0].name, "?0");
  ASSERT_EQ(norm_inferredAconstant.preconditions.nodes[1].parameters[0].name, "a");

  plansys2::Derived fd_better;
  fd_better.predicate = parser::pddl::fromStringPredicate("(inferred-FunctionDesign ?x)");
  fd_better.preconditions =
    parser::pddl::fromString("(exists (?y) (and (inferred-FdBetterUtility ?x ?y)))");
  auto fd_better_norm = fd_better.getNormalizedDerived();

  plansys2::Derived fd_better_inverse;
  fd_better_inverse.predicate = parser::pddl::fromStringPredicate("(inferred-FunctionDesign ?y)");
  fd_better_inverse.preconditions =
    parser::pddl::fromString("(exists (?x) (and (inferred-FdBetterUtility ?x ?y)))");
  auto fd_better_inverse_norm = fd_better_inverse.getNormalizedDerived();

[...]

PR #392, opened 2025-12-15
Fix ci 2

fmrico:
Hi,

Some tests are failing—these errors were not detected in the PRs. This PR tries to fix all of them.

@Rezenders , I have deactivated tests (#391)
@ajtudela `plansys2_support_py` was not tested in the CI (it wasn't in the workflow file), and you should re-enable checks after this PR.

I will extend the description, but it is mainly for fixing the CI


ajtudela:
I will check the tests again when its merged.

PR #398, opened 2026-03-11
Update CI to use container workers

Butakus:
There is an issue with apt-get in bare metal GitHub actions workers when using the [setup-ros action](https://github.com/ros-tooling/setup-ros#overview), so they recommend using the container workers.

This PR migrates the execution from bare metal workers to containers. It also removes the CI matrix setup because it had only one variant.

Additionally, I added a `workflow_dispatch` event so the CI actions can be manually trigger from Gitub's interface (this has to be migrated to the other branches for it to work there).

Please let me know if you are ok with the changes or if we should change anything else (like the scheduled runs for jazzy/kilted).

### DIFF
--- a/.github/workflows/foxy-devel.yaml
+++ b/.github/workflows/foxy-devel.yaml
@@ -7,27 +7,25 @@ on:
   push:
     branches:
       - foxy-devel
-      
+  workflow_dispatch:
+
 jobs:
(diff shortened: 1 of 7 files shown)


fmrico:
Merged :rocket: 

Thanks!!

Issue #397, opened 2026-03-11
Deprecated functions (warnings)


Butakus:
Hi,

When building the `rolling` version, I get warning messages about these deprecated functions:

```
warning: ‘std::string ament_index_cpp::get_package_share_directory(const std::string&)’ is deprecated: Use get_package_share_directory(..., std::filesystem::path) instead

...

warning: ‘void rclcpp::spin_some(Node::SharedPtr)’ is deprecated: use SingleThreadedExecutor::spin_some instead

...

warning: ‘plansys2::ActionExecutorClient::ActionExecutorClient(const std::string&, const std::chrono::nanoseconds&)’ is deprecated: Use ActionExecutorClient(const std::string & node_name) instead
```

The fix should be pretty straightforward if anyone wants to contribute. I may be able to take it myself in a few weeks if no one does it before.

zeel2104:
I can work on this issue. Can you assign it to me?

Butakus:
Sure!

Feel free to send the PR to the rolling branch whenever it's ready. We can use this issue if you encounter any problem.

Thank you in advance!

Issue #381, opened 2025-09-12
Duplicate node creation


jmilligan-launchpad:
Hello,

I've been working to integrate PlanSys2 (jazzy) with my own project , but I noticed that it was creating duplicate nodes for the problem, domain and planner clients. 

It looks to me as though the executor creates an instance of each of it's own for these. 

But it also seems to create a duplicate problem and domain node on each new plan execution. I'm curious if I'm doing something wrong or if it's an inherent issue.

I tested this with the examples: patrol_navigation and was seeing the same problem:

```
plansys2_node-1] [WARN] [1757687605.959491608] [rcl.logging_rosout]: Publisher already registered for node name: 'domain_expert_client'. If this is due to multiple nodes with the same name then all logs for the logger named 'domain_expert_client' will go out over the existing publisher. As soon as any node with that name is destructed it will unregister the publisher, preventing any further logs for that name from being published on the rosout topic.
[plansys2_node-1] [WARN] [1757687606.014899957] [rcl.logging_rosout]: Publisher already registered for node name: 'problem_expert_client'. If this is due to multiple nodes with the same name then all logs for the logger named 'problem_expert_client' will go out over the existing publisher. As soon as any node with that name is destructed it will unregister the publisher, preventing any further logs for that name from being published on the rosout topic.
```

and this is reflected by there being multiple listed when doing a ros2 node list:

```
/domain_expert
/domain_expert_client
/domain_expert_client
/domain_expert_client
/domain_expert_client
/domain_expert_lc_mngr
/executor
/executor_client
/executor_lc_mngr
/move_action_node
/nav2_sim_node
/patrol_action_node
/patrolling_controller
/pick
/planner
/planner_client
/planner_client
/planner_lc_mngr
/problem_expert
/problem_expert_client
/problem_expert_client
/problem_expert_client
/problem_expert_client
/problem_expert_lc_mngr
/screw
```

I'm curious if I'm doing something wrong that may be causing this. The code works fine, but creating so many duplicate nodes is obviously a problem and not just something I can ignore.

Any help is hugely appreciated, thank you.

fmrico:
It shouldn't. I will check it

PR #405, opened 2026-04-07
Rolling pr duplicated nodes names

SamueleSandrini:
Hi @fmrico,

while I am doing some tests I noticed an issue we had some time ago with multi-robot experiments related to the node name duplication issue described in #381, particularly affecting `domain_expert_client` and `problem_expert_client`.
In this pull request, I propose an approach to avoid the situation where multiple nodes have the same name.

image

## Problem
When client nodes are instantiated as class attributes, they inherit global arguments and end up sharing the same node name as the parent node. This leads to multiple nodes with identical names in the ROS graph, which can cause unexpected behavior.

## Proposed solution
This PR introduces two changes:

1. **Disable global arguments for internal nodes**  
   Both C++ and Python client implementations now explicitly disable global arguments when creating internal nodes, preventing unintended name reuse.

2. **Explicit naming for client nodes**  
   Added constructors that allow specifying node names.  
   Internal client nodes are now named using the pattern: `{external_node_name}_{internal_node_name}`. Let me know if this approach works for you, or if you would prefer `{internal_node_name}_{external_node_name}` to group all `domain_expert_client` / `problem_expert_client` nodes. I am also open to exploring alternative approaches if you think there are better ways to avoid this proliferation of duplicate nodes. I was also thinking about whether it might be a good idea to mark them as hidden nodes by using an underscore as a node name prefix, as they might be more confusing than helpful (view all `domain/problem_expert_client` nodes), let me know what you think.  
This ensures uniqueness and improves graph readability.

Example result:
- No duplicated node names
- Clear association between external node and its internal clients
Screenshot from 2026-04-07 11-44-44

## Additional changes
- Python parser: ignore comments inside PDDL expressions to avoid parsing issues
- ~~a minor adjustment in the use of `ament_index_cpp::get_package_share_directory` in `ComputeBT` and `ExecutorNode` to improve cross-compatibility between ROS versions (verified with Jazzy).~~  I removed this changes, as I noticed I re-introduces the warnings that was already addressed in https://github.com/PlanSys2/ros2_planning_system/issues/397#issue-4056957698. Nevertheless, that is the only aspect that makes the rolling release incompatible with Jazzy. What do you think about introducing a utility/wrapper function to make it cross-compatible?

Thanks!

### DIFF
--- a/plansys2_domain_expert/include/plansys2_domain_expert/DomainExpertClient.hpp
+++ b/plansys2_domain_expert/include/plansys2_domain_expert/DomainExpertClient.hpp
@@ -63,6 +63,13 @@ public:
   DomainExpertClient();
 
   /**
+   * @brief Construct a new DomainExpertClient object with a given node name.
+   *
+   * @param node_name Name of the ROS2 node.
+   */
+  DomainExpertClient(const std::string & node_name);
+
+  /**
    * @brief Get the name of the domain.
    *
    * @return std::string The name of the domain.
--- a/plansys2_domain_expert/include/plansys2_domain_expert/DomainExpertClient.hpp
+++ b/plansys2_domain_expert/include/plansys2_domain_expert/DomainExpertClient.hpp
@@ -67,7 +67,7 @@ public:
    *
    * @param node_name Name of the ROS2 node.
    */
-  DomainExpertClient(const std::string & node_name);
+  explicit DomainExpertClient(const std::string & node_name);
 
   /**
    * @brief Get the name of the domain.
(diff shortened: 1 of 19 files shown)


fmrico:
Hi @SamueleSandrini 

Thanks for detecting this and the work made here.

I'm hesitant to expose the internal node name as an argument in the client constructor. I believe this creates unnecessary coupling, as it's confusing for the user to understand that it's for the name of an auxiliary internal node we're creating. This unnecessarily increases unwanted coupling. I'd prefer not to expose it and instead create random node names. For example, all the nodes that create the `problem_clients` could start with the same name, but then have a random component, like "problem_client_askfljs" or something similar. What do you think?

PR #406, opened 2026-04-15
Support custom messages in BTAction via out_msg blackboard

avalen2022:
Read out_msg from the BT blackboard and use it as the status string in send_feedback() and finish() calls. Falls back to the default generic messages when out_msg is empty. This allows BT action nodes to forward custom context (e.g. perception events) through the action hub.


fmrico:
LGTM

Merging!!! :rocket:

PR #407, opened 2026-04-20
msgs: deal with CMake `CMP0094`

gavanderhoorn:
Bump CMake minimum version to `3.15`, as that's where `CMP0094` was introduced.

CMake versions lower than `3.15` have trouble locating Python in environments where a Python installed in a non-default location should be used (such as Conda, Pixi, etc).

### DIFF
--- a/plansys2_msgs/CMakeLists.txt
+++ b/plansys2_msgs/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.5)
+cmake_minimum_required(VERSION 3.15)
 
 project(plansys2_msgs)
 


gavanderhoorn:
I've only bumped the minimum version for `plansys2_msgs` as that's where I ran into this, but it might make sense to bump it in all packages to stay consistent.

And this should be possible to backport all the way back to Humble, as Humble (on Ubuntu) by default is build with CMake 3.22.1, so 3.15 should always be available.

gavanderhoorn:
Friendly ping @fmrico

fmrico:
Hi @gavanderhoorn 

Thanks for this contribution. Yes, it makes sense to bump the minimum version for all packages. If you want to do it in a separate PR, go ahead. If not, I can prepare a PR with these changes.

I will merge this :rocket:

gavanderhoorn:
@fmrico: apologies for not getting back to this.

I haven't found the time to bump versions on all other packages, but could we at least backport the change in this PR to other branches, like `humble-devel` and `jade-devel`?

PR #409, opened 2026-06-05
Fix deprecated get_share_directory

ajtudela:


fmrico:
Thanks @ajtudela :rocket:

